PR #71, opened 2015-01-03
reduce memory usage

shibukawa:
xlsx module has tree structure like this:

```
File -> Sheet -> Row -> Cell
Cell -> Style
```

And each child had reference to parent too, but Sheet -> File and Cell -> Row links didn't use pointer. Sheet and Cell structs had parents as instance. I think it is a bug (e.g. sheet1's parent and sheet2's parent are different instance). And each Cell object had Style instance. Now Cell type has Style pointer and cells share style instance. 

I made sample spreadsheet (it contains 1600 cells * 4 sheets) with following fix reduce memory usage 51% (3.22 MB -> 1.57MB). Performance improves 1.7%.

```go
// sample benchmark code
package main

import (
    "log"
    "time"
    "runtime"
    "github.com/tealeg/xlsx"
)

func main() {
    var book *xlsx.File
    start := time.Now()
    for i := 0; i < 100; i++ {
        book, _ = xlsx.OpenFile("bench.xlsx")
    }
    end := time.Now()
    log.Printf("sheets: %d\n", len(book.Sheets))
    log.Println("time: ", end.Sub(start).Nanoseconds() / 1000000, "ms")

    var memstats *runtime.MemStats
    memstats = new(runtime.MemStats)
    runtime.ReadMemStats(memstats)
    log.Printf("allocated memory: %d", memstats.Alloc)
}
```

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -20,10 +20,10 @@ const (
 // Cell is a high level structure intended to provide user access to
 // the contents of Cell within an xlsx.Row.
 type Cell struct {
-	Row      Row
+	Row      *Row
 	Value    string
 	formula  string
-	style    Style
+	style    *Style
 	numFmt   string
 	date1904 bool
 	Hidden   bool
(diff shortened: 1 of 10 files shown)


tealeg:
That's great!  Thanks for your work!

PR #72, opened 2015-01-05
error -> err

gyuho:
`error` is a type keyword in Go. Could you change it to `err`? And thanks for the package!


tealeg:
@gyuho good call.  Thanks for your patch.

gyuho:
Thank you for the great package!

PR #73, opened 2015-01-06
fix bug when loading spreadsheet (style cache map was accessed from many...

shibukawa:
... goroutines)

When loading spreadsheet that has many sheet, "fatal error: bad map state" was happened randomly (3/10 rate). The last my pull request introduce style cache shared in a workbook. But this cache was accessed from many goroutines of sheets at the same time and fatal error was happened.

This pull request addes read/write lock to suppress this issue.

Andi fix format by using go fmt.

### DIFF
--- a/xmlStyle.go
+++ b/xmlStyle.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


tealeg:
Thanks.  I wonder if we can't do something with channels instead, but this will do for now.

shibukawa:
Channels are useful for serialising data access and synchronising several goroutines. In this case, goroutines access to cache randomly and synchronising is not needed. And channel access is slower than locking.

Issue #78, opened 2015-01-16
Question: leading empty rows


shibukawa:
When reading xlsx file like the following CSV that has two empty rows. xlsx module returns just two rows that have values (sheet.Cell(0, 0).Value == "A3").

```txt
"", ""
"", ""
"A3", "B3"
"A4", "B4"
```

readRowsFromSheet(https://github.com/tealeg/xlsx/blob/master/lib.go#L349) function returns just line 3 and 4 (omit empty leading rows). And I can't get any information how many rows were omitted from result of OpenFile(). It is not good for handling cell range string (like A2:C5).

I think there are two options:

1. Add empty rows and cells into xlsx.Sheet.Rows to keep Cell(0, 0) always points A1 cell.
2. Add MinRow, MinCol property to xlsx.Sheet to enable to handle omitted information from client code.

I think no.1 is an easy solution to understand the behaviour of xlsx module but it will break backward compatibility.

If you decide the which is better, I will create pull request ;)

tealeg:
I think we should strive to make the interface of XLSX be as much like using a spreadsheet as possible, to that extent the behaviour of omitting blank rows is probably undesirable at this point.  Personally I'm OK with solution 1 - we already have to do something similar for cells.  

I look forward to your pull request, please make sure you include tests!

shibukawa:
#79 close this issue

PR #79, opened 2015-01-18
Fix for #78 (it doesn't omit empty rows and cols)

shibukawa:
* Change default behavior when reading empty cols and rows. now it returns empty cells. 
* And now it reads Col.Width (L399 of lib.go).


tealeg:
Cheers!

PR #83, opened 2015-01-29
Added functionality for writing a struct to a row

EricLagerg:
I needed this for a project of mine so I figured it'd be useful. It allows the writing of any struct (e.g. `type Person struct{}`) to a row, cell-by-cell.

This library doesn't have support for writing `int64` to a cell, so on a 32-bit platform it may overflow an `int64` member. I could submit a PR that adds a `SetInt64` function or just change `SetInt` to `int64`

### DIFF
--- a/write.go
+++ b/write.go
@@ -16,7 +16,7 @@ func (r *Row) WriteStruct(e interface{}, cols int) int {
 	}
 
 	n := v.NumField() // number of fields in struct
-	if cols < n && cols > 0 {
+	if cols < n {
 		n = cols
 	}
 
--- a/write.go
+++ b/write.go
@@ -1,0 +1,44 @@
+package xlsx
+
+import "reflect"
+
+// Writes a struct to row r. Accepts a pointer to struct type 'e',
+// and the number of columns to write, `cols`. Returns -1 if the 'e'
+// doesn't point to a struct, otherwise the number of columns written
+func (r *Row) WriteStruct(e interface{}, cols int) int {
+	if cols == 0 {
+		return cols
+	}
+
+	v := reflect.ValueOf(e).Elem()
+	if v.Kind() != reflect.Struct {
+		return -1 // bail if it's not a struct
+	}
+
+	n := v.NumField() // number of fields in struct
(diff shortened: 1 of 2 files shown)


tealeg:
That's great @EricLagerg - thanks for your work.

I'd opt for adding a SetInt64 because it wouldn't cause problems for existing code.

EricLagerg:
Okay. While I'm at it, is there anything else you want that you can think of off the top of your head? I have some extra free time today :smiley:

tealeg:
@EricLagerg  - Eric - I don't have a specific wish list - oddly enough I don't use this library at all any more.  

With regards to Int() - the answer is that you can call Int() on a cell with a float value, if you so wish.  It'd be equally appropriate to have that raise an error instead, but I suspect this way around was convenient for me at some point.

EricLagerg:
Ah. Are there any other xlsx libraries you use? I haven't found any I've liked.

I guess I assumed the user would do the casting instead of the casting being done in the library, since the latter would result in a cell's max value is capped at 9.223372036854776e+18 (which is only like 53 bits of precision).

I'll add Int64() which will use strconv.ParseInt(val, 10, 64) instead of ParseFloat()

tealeg:
@EricLagerg ha! No, if I were going to use an xlsx library, I guess I'd use the one I created :-)

The back story is that I created this library several years ago when I was working in a company with staff managing some data in Excel that I needed to process and deliver to an outside vendor.  I wrote this library to make that happen.  Shortly after that I was hired by Canonical (Ubuntu) and I got to retreat back into my Linux only existence and Excel files were no longer a part of my daily life.

I maintain this library today because people seem to use it, and, well, I think it's important to "give back".

Thanks for you efforts today, much appreciated.

EricLagerg:
Lucky you! I've the utmost respect for Windows, but I'd almost be willing to pay money to work with only Linux :stuck_out_tongue_winking_eye: 

You're definitely welcome, and I'll keep adding contributions if I end up writing anything else.

I have a Go app that has to create Excel files from a database so finding this library was a godsend (saves me from writing C!), so thank you!

PR #85, opened 2015-01-29
Fix a number of small issues with number format handling

sirwart:
I noticed that number formats weren't being unmarshalled correctly from the excel files I was using, and traced it back to the fact it's attributes weren't marked as attributes. Since Marshal is handwritten it was only affecting unmarshalling. Fixing this exposed some other issues in number format handling that was causing the tests to fail:

1. The number formats were kept in a global map, even though their ids
   are reused between different workbooks. This caused the tests to
   interfere with each other. I fixed this by making it a per-stylesheet
   which matches Excel's model.
2. The method makeXLSXStyleElements returns a nil initialized numFmt.
   Its id was 0 which conflicts with the builtin general number format.
   It was only used in 1 place (and it no-oped before because of the
   previous bug) so I just removed it.

### DIFF
--- a/cell_test.go
+++ b/cell_test.go
@@ -43,7 +43,7 @@ func (s *CellSuite) TestSetStyleWithFonts(c *C) {
 	style.Font = *font
 	cell.SetStyle(style)
 	style = cell.GetStyle()
-	_, xFont, _, _, _, _ := style.makeXLSXStyleElements()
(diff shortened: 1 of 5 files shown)


tealeg:
Thanks very much, that's great.  FWIW, the manual marshalling is a temporary hack that I plan to remove.

PR #86, opened 2015-01-29
added int64(), setint64(), and (*Row)WriteSlice() functions

EricLagerg:
Basically what we talked about in https://github.com/tealeg/xlsx/pull/83 just with an added function for writing slices.


tealeg:
@EricLagerg awesome.  Thanks for your efforts!

PR #87, opened 2015-02-01
Add support for the builtin number formats

sirwart:
Most of these I sourced from this list:
http://polymathprogrammer.com/2011/02/15/built-in-styles-for-excel-open-xml/

I found 41 through 44 in ECMA-376 Fourth Edition, Part 1, in
/OpenOfficeXML-SpreadsheetMLStyles/presetCellStyles.xml
(linked here:
https://www.dropbox.com/s/smvb305g6rfb6xj/presetCellStyles.xml)

### DIFF
--- a/xmlStyle.go
+++ b/xmlStyle.go
@@ -110,6 +110,75 @@ func (styles *xlsxStyleSheet) getStyle(styleIndex int) (style *Style) {
 
 }
 
+// Excel styles can reference number formats that are built-in, all of which
+// have an id less than 164. This is a possibly incomplete list comprised of as
+// many of them as I could find.
+func getBuiltinNumberFormat(numFmtId int) string {
+	switch numFmtId {
+	case 1:
+		return "0"
+	case 2:
+		return "0.00"
+	case 3:
+		return "#,##0"
+	case 4:
+		return "#,##0.00"
+	case 9:
+		return "0%"
+	case 10:
+		return "0.00%"
+	case 11:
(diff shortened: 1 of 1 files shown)


tealeg:
Ah yes, I'm actually aware of these, but thanks for the work in any case!

I have some ongoing work to better handle number formats in general.  Once I get a little time... ;-)

sirwart:
Just curious, what in particular were you thinking about? I thought it'd be relatively easy to add a more robust string formatting but it's a pretty hard problem to solve separate from rendering since they contain commands like centering, adding non-standard amounts of space, etc.

tealeg:
@sirwart I'm considering a full parser for the formatting language.  I did start some code in this direction, but I haven't touched it in a while.  I think I need to decide to do all or nothing with regards to formatting.

sirwart:
I'm going to be working on better formatting in the next few weeks, but there's a decent chance I push most of it out to the client (which in my case is a web browser). I wrote a pretty hax formatter function in Go which strips out all non-character info (like colors and whitespace), but I'm skeptical that I'll stick with that solution for long because it looks so much worse than how Excel renders it. I can see how the less robust solution could be useful for some people but it's not a clear win like some other parts of xlsx.

tealeg:
@sirwart Right, exactly - presenting the raw data from the file is clean, but only barely usable.  There will always be bugs until such time as a full implementation is completed.

Issue #63, opened 2014-11-25
Writer gives broken Excel File


pgundlach:
Using the following file with version d6607c512cf8c1083a9b5ec1efd7e1ccfb1832c8 I get a broken Excel file. How can I write multiple rows?

````
package main

import (
	"fmt"
	"github.com/tealeg/xlsx"
)

func main() {
	var file *xlsx.File
	var sheet *xlsx.Sheet
	var row *xlsx.Row
	var cell *xlsx.Cell
	var err error

	file = xlsx.NewFile()
	sheet = file.AddSheet("Sheet1")
	row = sheet.AddRow()
	cell = row.AddCell()
	cell.Value = "I am a cell!"

	row = sheet.AddRow()
	cell = row.AddCell()
	cell.Value = "I am a cell!"

	err = file.Save("MyXLSXFile.xlsx")

	if err != nil {
		fmt.Printf(err.Error())
	}
}
````

**Comment by kardianos** [comment]
@tealeg You can use MS One Drive (Free to use, can use from Chrome on Linux) to test XLSX output.

**Comment by tealeg** [comment]
Thanks @kardianos, I actually started doing just that over Christmas!  It's still not perfect, in that it doesn't tell me anything about what is wrong with the files, but at least it gives me *some* indication.

**Comment by BrianMMcClain** [comment]
FWIW, seeing the same on:

OS X 10.10.2
Microsoft Excel for Max 2011 v14.4.7 (141117)

And seeing the following logs after a repair

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<recoveryLog xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"><logFileName>Repair Result to test 06308.xml</logFileName><summary>Errors were detected in file 'Macintosh HD:Users:brianmcclain:Documents:test.xlsx'</summary><removedParts summary="Following is a list of removed parts:"><removedPart>Replaced Part: /xl/worksheets/sheet1.xml with XML error.  Load error. Line 2, column 698.</removedPart></removedParts><removedRecords summary="Following is a list of removed records:"><removedRecord>Removed Records: Document Theme from /xl/workbook.xml (Workbook)</removedRecord></removedRecords><repairedRecords summary="Following is a list of repairs:"><repairedRecord>Repaired Records: Format from /xl/styles.xml (Styles)</repairedRecord></repairedRecords></recoveryLog>
```

[...]

PR #88, opened 2015-02-03
Fix for broken Excel File

crahles:
I compared a broken file against a repaired version from Excel and could track it down to 2 problems:
- [Content_Types].xml must reference the xl/theme/theme1.xml file (commit crahles/xlsx@7d34e79)
- xl/styles.xml must include fill and border definitions, so i added them as defaults to NewStyle (commit crahles/xlsx@e0732a1)

I slightly expanded the tests to cover the changed behaviour and tested written files with LibreOffice, MS Office  2011 Mac and MS Office 2013 Win. Hope this helps on tealeg/xlsx#63.


style.go
```diff
@@ -90,3 +94,12 @@ func NewFont(size int, name string) *Font {
 func DefaulFont() *Font {
 	return NewFont(12, "Verdana")
 }
+
+func DefaulFill() *Fill {
+	return NewFill("none", "FFFFFFFF", "00000000")
+
+}
+
+func DefaulBoder() *Border {
```
tealeg: There's a typo here, and in every place it's used. DefaultBoder should be DefaultBorder.

tealeg:
Thanks for your work!

PR #89, opened 2015-02-03
Added SetColWidth Function

achernyak:
I am adding this function in order to allow for setting column width in your worksheet. It works for both ranges and single cells.


tealeg:
Thanks @achernyak

PR #90, opened 2015-02-05
Add basic support for reading freeze and split state out of worksheets.

sirwart:
I've gotten this far enough to thread the data through I care about successfully, but wanted to get your feedback on a couple of things.
1. How much of the sheet view do you want implemented before exposing it? Right now it just has the Pane object
2. Should SheetView and Pane be in their own files? They don't have any methods on them so it seemed like overkill but it's consistent with what you've done to other publicly exposed structs like Row
3. I tacked onto an existing test case. Should it be in a separate test case?

Thanks in advance

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -459,6 +459,28 @@ type indexedSheet struct {
 	Error error
 }
 
+func readSheetViews(xSheetViews xlsxSheetViews) []SheetView {
+	if xSheetViews.SheetView == nil || len(xSheetViews.SheetView) == 0 {
+		return nil
+	}
+	sheetViews := []SheetView{}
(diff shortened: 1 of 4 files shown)


tealeg:
Hi @sirwart 

1. I'm happy exposing this code as it is.  I don't mind partial support, so long as what's there is working.
2. I think they're OK where they are - if they grow unweildy then we can look at it.
3. What you've done is consistent with the project.  Technically its better to have an isolated test for every little thing (unit testing, in it's truest sense) and I'd be an advocate for that generally,, but that's quite hard to achieve when we're testing against complex inputs (XLSX files) so I've been somewhat "pragmatic" in my own approach and I expect nothing more from others..

PR #92, opened 2015-02-19
Fix: WriteSlice() will always return -1

yunkai:
Signed-off-by: Yunkai Zhang <qiushu.zyk@taobao.com>

### DIFF
--- a/write.go
+++ b/write.go
@@ -11,13 +11,13 @@ func (r *Row) WriteSlice(e interface{}, cols int) int {
 		return cols
 	}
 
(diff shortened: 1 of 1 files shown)


tealeg:
This patch causes panics in the test suite.  Please address that.

yunkai:
Sorry, I found that I misunderstood the WriteSlice() API.

When we call this function, we should pass address of slice -- not the slice itself.

So the code is correct and my patch is unnecessary.

But I think the API is not in Golang's idiomatic style:) --using slice as function argument is cheap.

PR #96, opened 2015-03-05
Default Style Option & Skip Blank Cells

mackross:


tealeg:
@mackross if you'd be so good as to add tests to cover these changes I'd be very happy to accept them.

mafredri:
@mackross how does this PR deal with empty cells that should be present? I can think of many scenarios where a blank cell is required to be present in the document.

mackross:
Setting a value of "" will cause the cell to write out to xml. Calling Reset() on the cell will toggle the write flag and stop the cell being written out. For large documents this represents a huge size reduction.

mafredri:
Wouldn't it be better to explicitly state that a cell can be removed if left empty? Say you add a cell that should merge with another cell, but that cell get's removed because it wasn't given a value, and thus merges with the wrong cell? I'm not saying that's what happens, just a concern!

PR #97, opened 2015-03-06
Fix number formats in workbooks where no custom formats are created

sirwart:
The map wasn't created until the first addFmt was called, which
isn't called for custom formats, which was preventing builtin
formats from being read.


tealeg:
Great, thanks for your work!

PR #100, opened 2015-03-15
Initial support for themes

sirwart:
Right now this is only used to color cells that use theme colors


tealeg:
@sirwart fantastic.  Thank you very much for you work!

sirwart:
No problem. It's been really good having such a great library to start with!

PR #105, opened 2015-04-17
correct text format mistakes in Cell.FormattedValue function

qiuly:
the cell's number format "@" suggests it should be treated as text, trying to call strconv.ParseFloat() on    it will cause the following error:
strconv.ParseFloat: parsing "xxx": invalid syntax


ShawnMilo:
This has been added, with tests, in the master branch. Thank you for the cleaner solution.

PR #108, opened 2015-04-23
Added bold/italic/underline style support for writing.

xStrom:
Some initial support for B/I/U when reading XLSX was already there, I went ahead and added writing support as well.

### DIFF
--- a/style.go
+++ b/style.go
@@ -35,6 +35,21 @@ func (style *Style) makeXLSXStyleElements() (xFont xlsxFont, xFill xlsxFill, xBo
 	xFont.Family.Val = strconv.Itoa(style.Font.Family)
 	xFont.Charset.Val = strconv.Itoa(style.Font.Charset)
 	xFont.Color.RGB = style.Font.Color
+	if style.Font.Bold {
+		xFont.B = &struct{}{}
+	} else {
+		xFont.B = nil
+	}
+	if style.Font.Italic {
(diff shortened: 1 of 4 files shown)


ShawnMilo:
I haven't dealt with any style stuff yet, so I'm going to leave this one for @tealeg to review.

tealeg:
Hi @xStrom - I don't quite understand why you're using an empty struct as the value for those fields.  Could you just add a brief explanatory comment in the code and then ping me to re-review?

@ShawnMilo  - the Style stuff is horrible at the XML level, it has custom marshaling right now, and it shouldn't have.  This comes from trying to work out issues with why various versions of office were falling over reading the files (and in particular, trying to add "omit if empty" logic that doesn't work out-of-the-box with Go's XML marshaling).

xStrom:
@tealeg I'm using empty struct pointers (and custom marshalling) because that was what the code already used. It's not something I chose. The struct pointers were added in #99, which you yourself merged it seems.

My goal was simply to add my desired feature without refactoring the existing code base. People usually want these sorts of changes to be in separate pull requests.

tealeg:
@xStrom - you're quite right!   I wasn't suggesting you added the custom marshaling, I was making a separate point to @ShawnMilo  that this part of the code is unnecessarily ugly right now (and that's my fault!)  

I'm happy to accept this patch as is.

Issue #91, opened 2015-02-18
Xlsx file created by this library can't be open by Mac Numbers


yunkai:
I created an xlsx file with the example code in [readme.org](https://github.com/tealeg/xlsx/blob/master/README.org).

I can open it with MS excel, but Mac Numbers can't open it.

Does this project not support compatibility with Mac Numbers?

drussel:
+1 I was just going to open such an issue. Google seems to not be entirely happy about them either: it can't preview them, but it can import them.

tealeg:
Sadly I don't have a Mac or Numbers to test with so I am in no position to
offer any guarantees of compatibility.

Writing files is still a very experimental feature and I'm happy that it
seem to be working with Excel and LibreOffice currently.  I'm sorry to say
that the time I'm able to spend on this project right now is tiny.

ShawnMilo:
What does "can't open it" mean -- what specific error message do you get? I currently have a Mac available. Please create three bare-minimum files (one sheet, one cell) will MS Excel, Numbers, and this project with the same data and post them for comparison. Any error messages you can get from Numbers might be helpful.

Also, could this be a duplicate of #63? Please check and let me know.

elgs:
@ShawnMilo, I had the same problem. I created a bare-minimum file using the sample code from the README.md page. MS Office opens it without any problem, when I double click the file on a Mac, Numbers says "“a.xlsx” can’t be opened.", where I save the file as a.xlsx. My Numbers version is 3.5.2 (2150). OS X version is: 10.10.3 (14D136).

ShawnMilo:
There's currently a workaround in #63. I'm waiting to heard something from @tealeg to see whether this could be the official fix.

ShawnMilo:
Closed by c1b6df96.

PR #111, opened 2015-05-11
Hack to fix #63 and #91.

### LINKED ISSUE #63: Writer gives broken Excel File

ShawnMilo:
Uses a string replacement to bypass current limitation on
encoding/xml.

Proposed for @tealeg to approve or veto.

### DIFF
--- a/file.go
+++ b/file.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"strconv"
+	"strings"
 )
 
 // File is a high level structure providing a slice of Sheet structs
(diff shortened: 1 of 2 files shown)


tealeg:
@ShawnMilo   I'm OK with this as a temporary fix.  The tests need to pass mind.

ShawnMilo:
Oops. I ran the tests, but evidently not after the latest change. Tests now pass, re-pushed. Once Travis says it's okay then I'll merge.

PR #112, opened 2015-05-13
Improve boolean check.

ShawnMilo:
Updates `Cell.Bool()` to more accurately return a boolean value based on the cell's type.

If it's a boolean cell type, return the boolean value.

If it's a numeric type, `False` for zero and `True` for anything else.

Otherwise, return `False` for `""` or `True` for anything else.


ShawnMilo:
@tealeg: Please review.

tealeg:
Good stuff.

PR #113, opened 2015-05-16
Fix a couple of bugs + default col width

sirwart:
I got xlsx working better with Excel files created by Google Sheets which generates slightly different xml for xSplit/ySplit and font attributes. I think fixing these only made the code more robust though. I also exposed default row height and col width.

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -6,8 +6,10 @@ import (
 	"strconv"
 )
 
+// CellType is an int type for storing metadata about the data type in the cell.
 type CellType int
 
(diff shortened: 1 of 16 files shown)


ShawnMilo:
I took a quick look. I like the clean up done (replacing `struct{}` with more explicit values). I see that the tests pass, but there are no new tests added to check anything to do with `SheetFormat`. Should there be?

sirwart:
Good point, I added some tests

PR #115, opened 2015-05-28
Added date format "yyyy/mm/dd"

klniu:
Write date to cell with the format "yyyy/mm/dd", such as 2006/01/02


tealeg:
@klniu Please add a test case to cover this.

PR #117, opened 2015-05-28
Fix for a nil formula when the raw cell type is error (e).

### LINKED ISSUE #114: invalid memory address or nil pointer dereference for big xlsx
Hello,

A big xlsx file that previously was loaded ok, after update of xlsx, is failing with this error.
Was any change made to code to limit the size of loadable files ?

Ricardo.

ShawnMilo:
This fixes the problem and tests all still pass. But is it
the correct solution? Can we get rid of the recover() function?

Fixes #114.

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -320,6 +320,9 @@ func formulaForCell(rawcell xlsxC, sharedFormulas map[int]sharedFormula) string 
 	var res string
 
 	f := rawcell.F
+	if f == nil {
+		return ""
+	}
 	if f.T == "shared" {
 		x, y, err := getCoordsFromCellIDString(rawcell.R)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ShawnMilo:
@tealeg: Please review.

tealeg:
@ShawnMilo I don't recall exactly, but I think the recover is only there to reduce the noise in the error  messages when using multiple go routines.  If it's causing problems then by all means remove it.  

With regards to the fix, what you're doing seems fine, I don't see any explicit encoding of the logic about the rawcell type being "e".  It might also be worthwhile to add that condition to a test case so we don't regress on this.

ShawnMilo:
@tealeg: Please review and merge, since I wrote this. Please refer to #114, this patch has been confirmed to fix the issue. If you don't feel the need to review my patches, just let me know. Otherwise, I'm going to keep giving you the veto power.

@abutre, FYI. :ok:

tealeg:
Good stuff @ShawnMilo.  FWIW I think it's good that we review each other's patches  I'll ask the same of you as and when I do some more work on XLSX ;-)

PR #118, opened 2015-05-30
Added support for writing float with NumFmt style.

klniu:
Added support for writing float with NumFmt style. Now it can write float into cell with number format but not the formatted value which sometimes cause the broken file warning in Microsoft Excel.

This patch writes float directly into cell, also write its number format into styles.xml. For example:

```go
cell.SetFloatWithFormat(37947.7500001, "h:mm:ss")
```
will generate:
```
-<c r="B1" s="1">
       <v>3.794775e+04</v>
</c>
```
and
```
<xf numFmtId="21" fontId="0" fillId="0" borderId="0" applyProtection="0" applyNumberFormat="1" applyFill="0" applyFont="0" applyBorder="0" applyAlignment="0">
<alignment wrapText="0" vertical="" textRotation="0" shrinkToFit="0" indent="0" horizontal=""/>
</xf>
```

WIth user-define number format, it also add numFmt as below:
```
<numFmt formatCode="yyyy/mm/dd" numFmtId="164"/>
```


sheet.go
```diff
@@ -110,6 +110,12 @@ func (s *Sheet) makeXLSXSheet(refTable *RefTable, styles *xlsxStyleSheet) *xlsxW
 				xFont, xFill, xBorder, xCellStyleXf, xCellXf := style.makeXLSXStyleElements()
 				fontId := styles.addFont(xFont)
 				fillId := styles.addFill(xFill)
+				// generate NumFmtId
+				xNumFmt := styles.newNumFmt(cell.numFmt)
+				// the NumFmtId of NumFmt user defined is greater than 164
+				if xNumFmt.NumFmtId >= 164 {
```
ShawnMilo: What's the "magic number" 164?

sheet_test.go
```diff
@@ -49,6 +49,53 @@ func (s *SheetSuite) TestMakeXLSXSheetFromRows(c *C) {
...
+
+	c.Assert(styles.CellXfs.Count, Equals, 4)
+	c.Assert(styles.CellXfs.Xf[0].NumFmtId, Equals, 0)
+	c.Assert(styles.CellXfs.Xf[1].NumFmtId, Equals, 1)
+	c.Assert(styles.CellXfs.Xf[2].NumFmtId, Equals, 14)
+	c.Assert(styles.CellXfs.Xf[3].NumFmtId, Equals, 164)
+	c.Assert(styles.NumFmts.Count, Equals, 1)
+	c.Assert(styles.NumFmts.NumFmt[0].NumFmtId, Equals, 164)
+	c.Assert(styles.NumFmts.NumFmt[0].FormatCode, Equals, "hh:mm:ss")
+
```
ShawnMilo: It seems to me that these tests should also make some assertions about the `FormattedValue()` of the cells. Does that make sense here?

xmlStyle.go
```diff
@@ -15,6 +15,44 @@ import (
 	"sync"
 )
 
+// Excel styles can reference number formats that are built-in, all of which
+// have an id less than 164. This is a possibly incomplete list comprised of as
+// many of them as I could find.
+var builtInNumFmt map[int]string = map[int]string{
+	0:  "general",
```
ShawnMilo: the first `map[int]string` is redundant and can be skipped.

[example](http://play.golang.org/p/BzV7aRSqzb)

xmlStyle.go
```diff
@@ -137,71 +176,12 @@ func (styles *xlsxStyleSheet) argbValue(color xlsxColor) string {
...
-		return "##0.0E+0"
-	case 49:
-		return "@"
-	}
-	return ""
+	formatCode, ok := builtInNumFmt[numFmtId]
+	if ok {
+		return formatCode
+	} else {
+		return ""
```
ShawnMilo: The map will return `""` if the value is missing, so this section could be shortened to just:

```go
return builtInNumFmt[numFmtId]
```

xmlStyle.go
```diff
@@ -291,7 +271,39 @@ func (styles *xlsxStyleSheet) addCellXf(xCellXf xlsxXf) (index int) {
...
 }
 
+// newNumFmt generate a xlsxNumFmt according the format code. When the FormatCode is built in, it will return a xlsxNumFmt with the NumFmtId defined in ECMA document, otherwise it will generate a new NumFmtId greater than 164.
+func (styles *xlsxStyleSheet) newNumFmt(formatCode string) xlsxNumFmt {
+	if formatCode == "" {
+		return xlsxNumFmt{NumFmtId: 0, FormatCode: "general"}
+	}
+	// built in NumFmts in xmlStyle.go, traverse from the const.
+	numFmts := make(map[string]int)
+	for k, v := range builtInNumFmt {
```
ShawnMilo: Why is `numFmts` duplicating `builtInNumFmt` instead of just using it?

xmlStyle.go
```diff
@@ -291,7 +271,39 @@ func (styles *xlsxStyleSheet) addCellXf(xCellXf xlsxXf) (index int) {
...
+	numFmtId = 164
+	for {
+		// get a unused NumFmtId
+		if _, ok = styles.numFmtRefTable[numFmtId]; ok {
+			numFmtId += 1
+		} else {
+			break
+		}
+	}
+	return xlsxNumFmt{NumFmtId: numFmtId, FormatCode: formatCode}
```
ShawnMilo: This doesn't look safe for concurrency; multiple simultaneous calls to this will result in the same integer for different styles. In fact, since the integer isn't stored anywhere, even synchronous calls will result in the same number. Does this hurt anything?

ShawnMilo:
@klniu: Thanks for the work! Overall it seems fine. I added some notes to be addressed. Thanks for adding tests with your initial submission as well; it helps to see them pass.

sheet.go
klniu: The number format id less than 164 is builtin. 
d4320de787cb2e14773b63537a50ea865e585c1d fixed this.

sheet_test.go
klniu: The patch only get FormatCode from cell but not write to cell, so I think TestFormattedValue is sufficient.

xmlStyle.go
klniu: fixed.

xmlStyle.go
klniu: The map was traversed to get NumFormatId according the FormatCode.

xmlStyle.go
klniu: FIxed in d4320de787cb2e14773b63537a50ea865e585c1d by adding sync lock. At the same time, the newNumFmt will also add the return xlsxNumFmt into styles.

klniu:
824b36b74e742217be506388acb492d2631f09de will fix the SetFloatWithFormat not writing the float with the original precision.

ShawnMilo:
@klniu: If you've addressed everything, please change the tag back to "needs review." If you can't change the tag, please let me know and I'll do it. I just don't want to review if you're still tinkering with it.

klniu:
Finished. Please review.

PR #119, opened 2015-05-30
Prevent workbooks with chartsheets from causing panics

sirwart:
Chartsheets are in the sheets listing but are not included in the file
map. Before, this would cause a panic when an attempt is made to open
a nil file. Now, chartsheets are skipped until there's better support
for them.


ShawnMilo:
@sirwart: Thanks for the patch.

Please add a test which consumes a small sample file to prevent regressions. There are already some small files in `testdocs`, so maybe add one there.

I'd also like to see that file so I can reproduce the error and understand the fix better.

Thanks.

sirwart:
I added a test workbook (testdocs/testchartsheet.xlsx) and a test that verifies that it can be opened. If you run that test on master without my earlier patch it will fail

ShawnMilo:
Great, thanks. I'll take a look.

PR #121, opened 2015-06-03
Added ability to set the row height.

greenenergy:


tealeg:
@greenenergy Hi, please can you add unit test code for this change so we can ensure it continues to work in the future.

row.go
```diff
@@ -1,9 +1,16 @@
...
 
 type Row struct {
-	Cells  []*Cell
-	Hidden bool
-	Sheet  *Sheet
+	Cells    []*Cell
+	Hidden   bool
+	Sheet    *Sheet
+	Height   float64
+	IsCustom bool
```
ShawnMilo: `IsCustom` is about `Height`, but the name doesn't make that clear. Maybe `CustomHeight` as done in `xlsxRow`? I like the `Is` prefix (maybe even make it `is`, since it doesn't have to be shared outside the package.

sheet.go
```diff
@@ -110,6 +110,10 @@ func (s *Sheet) makeXLSXSheet(refTable *RefTable, styles *xlsxStyleSheet) *xlsxW
 		}
 		xRow := xlsxRow{}
 		xRow.R = r + 1
+		if row.IsCustom {
+			xRow.CustomHeight = true
+			xRow.Ht = fmt.Sprintf("%g", row.Height)
```
ShawnMilo: It might be nicer to just set `xRow.Ht = row.Height()`, and add
```go
`func (r *Row) Height() float64 {
    if r.CustomHeight {
        return fmt.Sprintf("%g", r.Height)
    }
    return fmt.Sprintf("%g", <default here>)
}
```
It might even be better to eliminate `CustomHeight` entirely, and just check whether `r.Height` is zero here. 

It would be ideal to have a default height set, and automatically set it upon row creation with `.AddRow`, but rows are created in multiple places, so I don't know how much work it would be to ensure every place gets the default. So just checking for zero here would be simpler. Any thoughts, @tealeg?

greenenergy: Well, the way I have it, it only adds a height specifier if the user changes it from the default, thus making a simpler XML file (since only "non-default" values get written to the file).

If you do it the way you propose, then every row gets a row height explicitly, and now you have to go look up what the default was to determine if it's been changed.

Anyway - I tend towards a "less is more" approach, so my tendency would be to let defaults be defaults and only override explicitly when necessary, but if you guys want it the other way, that's fine by me.

ShawnMilo: Yeah, that does make sense. I still think the boolean value can go away in favor of checking whether the custom height is zero. Less _is_ more.

greenenergy: Ok - I can change it to use the Height value and drop the bool.

tealeg:
@greenenergy cheers!  Merged.

PR #122, opened 2015-06-07
slightly improves test coverage

ShawnMilo:
Better coverage of `Cell.FormattedValue` in preparation for trying to dynamically handle more format strings.

### DIFF
--- a/cell_test.go
+++ b/cell_test.go
@@ -125,6 +125,9 @@ func (l *CellSuite) TestFormattedValue(c *C) {
 	// non-comma form.
 	c.Assert(cell.FormattedValue(), Equals, "37947")
 
+	cell.numFmt = "#,##0.00;(#,##0.00)"
+	c.Assert(cell.FormattedValue(), Equals, "37947.75")
+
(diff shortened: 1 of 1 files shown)


cell_test.go
```diff
@@ -143,6 +146,9 @@ func (l *CellSuite) TestFormattedValue(c *C) {
 	negativeCell.numFmt = "#,##0 ;[red](#,##0)"
 	c.Assert(negativeCell.FormattedValue(), Equals, "(37947)")
 
+	negativeCell.numFmt = "#,##0.00;(#,##0.00)"
+	c.Assert(negativeCell.FormattedValue(), Equals, "(-37947.75)")
```
ShawnMilo: This matches the actual output of the code. However, based on [this](https://github.com/tealeg/xlsx/blob/master/cell.go#L236), I suspect the code is wrong, and should use `Abs()` to remove the minus sign.

@tealeg

ShawnMilo:
Ignoring this in favor of #123.

PR #124, opened 2015-06-18
properly check column hidden for cell hidden logic

trinchan:
There is a bug with hidden cells introduced in this commit.
https://github.com/tealeg/xlsx/commit/00de551dc6de154cfa6db6b102a41448b251b8ad#diff-1861a0c40780efdec0f4e6e2e19b3d44R429

```diff
-			row.Cells[cellX].date1904 = file.Date1904
-			row.Cells[cellX].Hidden = rawrow.Hidden || (len(cols) > cellX && cols[cellX].Hidden)
+			cell.date1904 = file.Date1904
+			cell.Hidden = rawrow.Hidden || (len(cols) > cellX && cell.Hidden)
```

`cell` is set to `row.Cells[cellX]` so replacing `cols[cellX].Hidden` with `cell.Hidden` is incorrect. I changed it back to `cols[cellX].Hidden` so that the column hidden status is properly checked. I also added a test case for this.

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -508,7 +508,8 @@ func readRowsFromSheet(Worksheet *xlsxWorksheet, file *File) ([]*Row, []*Col, in
 				cell.numFmt = file.styles.getNumberFormat(rawcell.S)
 			}
 			cell.date1904 = file.Date1904
-			cell.Hidden = rawrow.Hidden || (len(cols) > cellX && cell.Hidden)
+			// Cell is considered hidden if the row or the column of this cell is hidden
+			cell.Hidden = rawrow.Hidden || (len(cols) > cellX && cols[cellX].Hidden)
 			insertColIndex++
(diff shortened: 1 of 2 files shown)


marexandre:
LGTM :+1:

ShawnMilo:
This looks good to me, but I'm going to ping @tealeg on it, since it touches more of the internals of the package than I have been exposed to.

Thanks for adding a test! My only comment is that I might change the text "hidden" inside the cell to "this is invisible" or something, because "hidden" is used as plain text, an XML value, and a property of a struct all in the same scope, which is a little confusing.

trinchan:
@ShawnMilo thanks, i updated the test content to be more clear

tealeg:
@trinchan really good spot, thanks!
@ShawnMilo nice review point.

PR #126, opened 2015-06-20
Fix panic on invalid value for "r" attribute in <row>.

### LINKED ISSUE #125: Unrecoverable panic on open file
I have a web application, and there is a panic/recover middleware to handle any type of un-handled panic, but here in this line : https://github.com/tealeg/xlsx/blob/b004da14c3ea7a710f0c8a4ee802af89c5952c98/lib.go#L611 there is a go rutine, and when I have a panic in read file (currently an index out of range) since its in another routine than my web request, it crash the app, and there is no way to handle this. 

Why this part is in a go routine??

ShawnMilo:
The value of the "r" attribute of a <row> was being used as part of a test
in a loop which loaded default values into a row. However, the value in
the attribute could be higher than the number of rows, causing a panic
due to an index error.

fixes #125

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -471,13 +471,17 @@ func readRowsFromSheet(Worksheet *xlsxWorksheet, file *File) ([]*Row, []*Col, in
 		rows[rowIndex] = makeEmptyRow()
 	}
 
+	numRows := len(rows)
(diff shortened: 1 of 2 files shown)


ShawnMilo:
@tealeg: Please review.

fzerorubigd:
Its not fixing #125 , its just fix another bug, the #125 is not about this particular bug.

PR #128, opened 2015-06-23
Adds a test the randomly modifies parts of working xlsx documents

HeinOldewage:
The library is given a slightly modified version of testfile.xlsx and tested for uncaught panics.

There are a few problems with the test, but I see no way around them:
The library panics in a goroutine and terminates the test, without the test every printing logged output, thus printing information to stdout as the test proceeds is the only way of getting output. I suppressed output unless the `-test.v` flag is set. The test always prints the random seed (the only line alway printed) that is is using in order for the test to be repeatable should it fail(set using `-test.seed`).
Thus when running the test the seed will be output, if it it fails the test can be rerun with the `-test.v` and `-test.seed=<seed>` flags to see what modification to the file broke the library. 

The test currently modifies tags and attributes to random strings and random integers(if the token was an integer before the modification). More modifications can be added to getTokenVariations.

The test fails due to a few different issues, some of them will require a significant rewrite of logic of some parts. I will briefly list the ones I have discovered:
* lib.go:390: the panic goes uncaught. My suggestion would be to return an error as the calling function might want to close a channel.

* reftable.go:55: ResolveSharedString might not be able to resolve an index, but this is not reported/handled. fillCellData might want to return a reasonable error should this occur.

* lib.go:609: The reader closes the channel, where the writer should. This causes writes on closed channels if an error is encountered. Moving the `defer close(sheetChan)` inside the goroutine on line 611 will solve this problem. The deferred function can then also set err, allowing readSheetsFromZipFile to correctly reporting errors. The loop starting on 618 is incorrectly stated, no assumption should be made as to the number of values written to the channel as the writer can panic. Changing it to `for sheet := range sheetChan{}` will correctly loop over the channel.

* lib.go:168: getCoordsFromCellIDString is far too graceful in the values it parses. (it parses MEbdVxTTqE6eebrPCMPNIzmF5Xx5JPup8Q35ipbw2OK without an error)
*lib.go:442: rowCount and colCount is 'inaccurate', if only one cell at ZZ26 has a value all cell/rows from A1 will be allocated. The rest of the library depends on those rows/cols existing even though they do not contain information. This mostly just inefficient, but a sheet crafted with a singular cell at a very large index can crash the library (make panics with makeslice: len out of range ).

* An error not tested for by this test, is missing files. Some files are expected and the library panics fatally is the do not exist.

Any advice around simpler testing setup and test output would be appreciated.

### DIFF
--- a/fuzzy_test.go
+++ b/fuzzy_test.go
@@ -1,0 +1,241 @@
+package xlsx
+
+import (
+	"archive/zip"
+	"bytes"
+	"encoding/xml"
+	"flag"
+	"fmt"
+	. "gopkg.in/check.v1"
+	"io"
+	"log"
+	"math/rand"
+	"path/filepath"
+	"reflect"
+	"strconv"
+	"testing"
+	"time"
+)
+
+type Fuzzy struct{}
+
+var _ = Suite(&Fuzzy{})
+var randseed *int64 = flag.Int64("test.seed", time.Now().Unix(), "Set the random seed of the test for repeatable results")
+
+type tokenchange struct {
+	file bytes.Buffer
+	old  xml.Token
+	new  xml.Token
+}
+
+type filechange struct {
+	File *zip.Reader
+	Name string
+	Old  xml.Token
+	New  xml.Token
+}
+
+var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
+var numbers = []rune("0123456789")
+
+func randString(n int) []byte {
+	b := make([]rune, n)
+	for i := range b {
+		b[i] = letters[rand.Intn(len(letters))]
+	}
+	return []byte(string(b))
+}
+
+func randInt(n int) []byte {
+	b := make([]rune, n)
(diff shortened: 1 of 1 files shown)


ShawnMilo:
Thanks for posting this. I'll look at it when I have time (probably the weekend), unless @tealeg gets to it first. I think this is very valuable considering it already found unreported bugs. Awesome.

tealeg:
It's nice to have a shakedown of how stuff fails.  Perhaps the best we can hope for from something like this is to tune our output in failure conditions so that they can be more easily diagnosed.

tealeg:
@HeinOldewage BTW - this is still on my radar - you've not been forgotten!

tealeg:
@HeinOldewage @ShawnMilo  - so, in order to sensibly merge this without breaking tests all the time I've added a build tag to the top of the fuzzy_test.go file.   This means that normal runs of "go test" will exclude the fuzzy_test.go file, if you wish to run it you can invoke:

    go test -tags=fuzzy

In general I suggest we proceed by creating issues for the problems fuzzy_test generates.

PR #132, opened 2015-06-30
Add Style & Type to Col

blackss2:
add default style to Col
add Col.SetType(cellType CellType)
add sheet.Col(idx int)
remove default style from Cell
Cell will be follow Col's style if don't have any style
add read col style from xlsx file
add SetDefaultFont(limitation - xlsx global setting)


tealeg:
Cheers @blackss2

PR #138, opened 2015-07-07
Add support for border colors

mafredri:
Hey, I needed the ability to apply colors to borders, so I made a quick patch. I would very much like to see this feature officially supported in this library, what are your thoughts on the implementation?

I'm open to making any changes you deem necessary.


tealeg:
That's great @mafredri - thanks for your work.

PR #140, opened 2015-07-10
get parsing error string when fetching the formated value for a wrong format cell

klniu:
Fix: get parsing error string when fetching the formated value for a cell with a wrong format.

In office, the cell whose value is a string or other type value can be set with a improper format. When parsing these value with the wrong format, FormattedValue will return the error string but not the value, such as: "strconv.ParseFloat: parsing \"string\": invalid syntax".

I think it's a improper behavior, so I commit this patch which will return the origin value when failing to parse.

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -278,7 +278,7 @@ func (c *Cell) GetNumberFormat() string {
 func (c *Cell) formatToFloat(format string) string {
 	f, err := strconv.ParseFloat(c.Value, 64)
 	if err != nil {
-		return err.Error()
+		return c.Value
 	}
 	return fmt.Sprintf(format, f)
 }
@@ -286,7 +286,7 @@ func (c *Cell) formatToFloat(format string) string {
 func (c *Cell) formatToInt(format string) string {
 	f, err := strconv.ParseFloat(c.Value, 64)
 	if err != nil {
-		return err.Error()
+		return c.Value
 	}
 	return fmt.Sprintf(format, int(f))
 }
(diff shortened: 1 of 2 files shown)


tealeg:
Sorry @klniu but I disagree - this is an error condition, and the decision about what to do about it should be made by the code at the call site in each consuming application, not by the library.  It is not the job of this library to unilaterally decide how to deal with bad data.

HeinOldewage:
@tealeg I agree that the calling application should decide how to deal with a parse error as opposed to the library and that this solution/pull request does not solve the problem.
However that is not currently possible to detect a parse error, in code, without parsing the output of `GetNumberFormat()`.

Ideally `GetNumberFormat()` should return a string and an error, but I assume we would like to avoid changing the API greatly?
The only other solution is to have `GetNumberFormat()` panic with a nice error. Something like:
`panic(fmt.Errorf("Could not format cell, %v",err))`
Perhaps a new pull request could be made with a better solution.

tealeg:
@HeinOldewage so, the current FormattedValue() command does what a spreadsheet would do (write an error instead of a value), however this is not very idiomatic for Go (and pretty useless as a programmer), so I propose instead we add a SafeFormattedValue function that does return the error, implement FormattedValue in terms of SafeFormattedValue and move on.

tealeg:
Well.. I accidentally committed my suggestion on master instead of on my branch (*doh* - lesson for the day - be careful working with multiple environments), so the above mentioned SafeFormattedValue is there now.  If anyone has objections let me know.

@ShawnMilo FYI.

PR #142, opened 2015-07-13
Apply customWidth attribute to cols and enable changing defaults

mafredri:
In Microsoft Excel for Mac 2011 width setting were not applied to columns. The attribute `customWidth` has to be set to `1` when a user has changed it ([DocumentFormat.OpenXml.Spreadsheet.Column](https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.column(v=office.14).aspx)). This has been remedied in this patch and the `customWidth` attribute is only applied when a user has explicitly changed the width of a column.

Previously default width was automatically applied during `maybeAddCol`, I removed it from here since it gets applied anyway in `makeXLSXSheet`. That way the we can check if `customWidth` need to be applied or not.

I also enabled modifications to `DefaultColWidth` and `DefaultRowHeight`. Now DefaultColWidth can be used instead of modifying each column separately.

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -110,6 +110,11 @@ func (s *Sheet) makeXLSXSheet(refTable *RefTable, styles *xlsxStyleSheet) *xlsxW
 	maxRow := 0
 	maxCell := 0
 
+	if s.SheetFormat.DefaultRowHeight != 0 {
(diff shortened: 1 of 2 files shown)


tealeg:
Great!  I've added in some unit tests for this behaviour.

PR #146, opened 2015-07-15
Set Sheet attr of each Row in readSheetFromFile

jangler:
Should be self-explanatory. Otherwise doing things like `row.AddCell()` on a file you just read can panic on a nil pointer dereference.


tealeg:
@jangler thanks for your MP, its a good find.

I reworked your changes so that Row.Sheet gets set everywhere we create a Row, this allowed me to avoid the extra loop in exchange for passing the Sheet around a little.  I also updated the unit tests and made assertions that the Row.Sheet was correctly set.  As a rule before you make a change you should ensure that there is at least one unit test in existence that fails prior to the change - the presence of such a test means we can avoid regressions if we restructure things in the future.

jangler:
Thanks @tealeg, I'll keep that in mind.

PR #149, opened 2015-07-22
Added merge cell borders

greenenergy:
We originally had both custom column width and merge cell borders in this patch, but coincidentally someone else did a custom column width patch, so we had to deal with that conflict.


mafredri:
Am I reading the changes wrong, or is this PR actually applying the same predefined style to all merged cells? Some concerns:

* `Rows[0].Cells[0]` is read, is it guaranteed to exist?
* Style is taken from `Rows[0].Cells[0]`, but that is not guaranteed to be correct for future merged cells
* Say you are merging Cells `0, 1, 2`. On cell 0 you apply a left border, on cell 2 you apply a right border, will this PR overwrite those changes?
* Style merging should perhaps be optional and the style needs to be taken from the cell containing `HMerge` / `VMerge`

syltaxue:
@mafredri 
Thanks for your feedback.

You are right, the changes are a bit unsafe for general uses.

The idea of reading  Rows[0].Cells[0] will be changed, and all it is doing is to create a pre-defined style for merged cells.

For merged cell styles, let me give you an example to explain how it works:

ShawnMilo:
I have marked this PR "in progress" until the merge conflicts and issues raised in the review my @mafredri have been addressed.

greenenergy:
Regarding @mafredri 's comments:
* For the first comment -- what code are you referring to? Can you make a comment in the context of the code itself?
* It's not? If not - what's the solution? What is your understanding of what's supposed to happen?
* Yes - according to the standard, when you merge cells, the first cell's styling wins and overrides the other cells, so the code is correct.
* This is a suggestion, and not even valid since the standard does not say anything about optional style merging (unless I missed it somewhere. Can you provide a link?)

In other words - as far as I can see, there are no actual outstanding issues that need to be addressed.

mafredri:
> For the first comment -- what code are you referring to? Can you make a comment in the context of the code itself?

@syltaxue already confirmed that it was unsafe use, but here goes.

```go
package main

import "github.com/tealeg/xlsx"

func main() {
	file := xlsx.NewFile()
	_, err := file.AddSheet("Sheet1")
	if err != nil {
		panic(err.Error())
	}

	err = file.Save("./foo.xlsx")
}
```

Result:

```
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/tealeg/xlsx.(*Sheet).makeXLSXSheet(0xc82009c000, 0xc82003f8d8, 0xc820076ea0, 0x0)
	/Users/mafredri/Golang/src/github.com/tealeg/xlsx/sheet.go:114 +0x26f9
github.com/tealeg/xlsx.(*File).MarshallParts(0xc82001a180, 0xe408, 0x0, 0x0)
	/Users/mafredri/Golang/src/github.com/tealeg/xlsx/file.go:227 +0x4cd
github.com/tealeg/xlsx.(*File).Write(0xc82001a180, 0x498210, 0xc82002c080, 0x0, 0x0)
	/Users/mafredri/Golang/src/github.com/tealeg/xlsx/file.go:112 +0x47
github.com/tealeg/xlsx.(*File).Save(0xc82001a180, 0x1ac010, 0xa, 0x0, 0x0)
	/Users/mafredri/Golang/src/github.com/tealeg/xlsx/file.go:99 +0xa0
main.main()
	/Users/mafredri/Golang/src/github.com/mafredri/xl/main.go:12 +0x235
```

> It's not? If not - what's the solution? What is your understanding of what's supposed to happen?

Yes it is. The master style is taken from the first cell and applied all merged cells which in my opinion is incorrect behavior.

```go
package main

import "github.com/tealeg/xlsx"

func main() {
	file := xlsx.NewFile()
	sheet, err := file.AddSheet("Sheet1")
	if err != nil {
		panic(err.Error())
	}

	row := sheet.AddRow()
	cell := row.AddCell()

	style := cell.GetStyle()
	style.Border.Top = "medium"
	style.Border.Bottom = "medium"
	style.Border.Left = "medium"
	style.Border.Right = "medium"
	style.ApplyBorder = true

	_ = row.AddCell()
	_ = row.AddCell()

	row = sheet.AddRow()
	_ = row.AddCell()
	_ = row.AddCell()
	_ = row.AddCell()

	row = sheet.AddRow()
	cell = row.AddCell()

	cell.Merge(1, 0)

	_ = row.AddCell()
	_ = row.AddCell()

	err = file.Save("./foo.xlsx")
}
```



I expect the cell(s) on row 3 to have no borders at all, but borders appear magically because of this logic.

> Yes - according to the standard, when you merge cells, the first cell's styling wins and overrides the other cells, so the code is correct.

Where does it say this in the standard? I would appreciated if you could point me in the right direction. I'm basing this off on normal Excel behavior and what I as a user would expect to happen.

#### Border behavior when merging

Before merge:



After merge (B2 and C2):



Borders are not inherited in the merge and cells keep the edge borders they have defined, only if borders are connecting are they kept:

#### Edge borders

Before:



After:



#### Keep connecting borders:

Before:



After:



> This is a suggestion, and not even valid since the standard does not say anything about optional style merging (unless I missed it somewhere. Can you provide a link?)

You are correct, I did not base this on any specification. I simply did not see how to satisfy user expectations with the implementation, which is why I suggested making it optional. The second part still stands though, the style cannot be assumed as it is now (already addressed by previous points).

greenenergy:
Thank you for your very thorough response, @mafredri . I'm on it now.

greenenergy:
@mafredri When I said "first cell", I meant the first cell in a merged group, not the first cell in the sheet. I agree with you - the current behaviour is incorrect.

greenenergy:
Regarding cell merge behaviour - Google Sheets and Open Office both take the styles of the upper-left cell in the group and use that for the merged group - in effect, the upper left cell expands to be the group.
Excel behaves as you describe, which I think is less useful and more surprising to the user than the google/openoffice technique, so I'm going to implement it in the google/openoffice manner.

It ultimately shouldn't matter that much, since the one consistent way to get borders and styles on a merged cell is to apply styles after the merge rather than before, as this works the same in all sheets.

mafredri:
> When I said "first cell", I meant the first cell in a merged group, not the first cell in the sheet. I agree with you - the current behaviour is incorrect.

That makes a lot more sense indeed. I've been thinking about this a bit and I can't really say I'm a fan of the implicit behavior of the first cell being the master cell for styling. But the only alternative I can think of is exposing a new kind of API, e.g. `mergedCell, err := xlsx.MergeCells(cell1, cell2)`. Now `mergedCell` could be styled and the styles of the actual underlying cells would be ignored.

PS. I saw you new reply as well re: gsheets and ooo, and I agree, it's probably more useful than the excel behavior. I still think my suggestion above might be worth considering. Thoughts?

mafredri:
Actually, I think my suggestion would require a lot of rethinking, especially with regard to reading in xlsx documents and making modifications to them (which cell is actually returned at row X col X, etc), for now it would probably just be more confusing than properly explaining the "first merged cell"-method.

greenenergy:
It's certainly not a bad idea, but I think I'd still like to try to emulate ooo/gsheets as much as possible, since it's reasonable and people are familiar with it already. I'm a big fan of leveraging  existing knowledge.
Ah, I see you beat me to it. :)

greenenergy:
Actually, don't merge this yet. The border issue is fixed for google sheets and open office, but Excel still has one last issue with it, which I will address now.

greenenergy:
Ok, this is ready for merge.

sheet.go
```diff
@@ -78,10 +78,22 @@ func (s *Sheet) Col(idx int) *Col {
...
+	r := sh.Rows[row]
+
+	if len(r.Cells) < (col + 1) {
+		numneeded := (col - len(r.Cells)) + 1
+		for x := 0; x < numneeded; x++ {
+			r.AddCell()
+		}
 	}
-	return new(Cell)
+	return r.Cells[col]
```
mafredri: I actually use something similar, but I believe this could read a bit easier like so:

```golang
for len(sh.Rows) <= row {
	sh.AddRow()
}
r := sh.Rows[row]
for len(r.Cells) <= col {
	r.AddCell()
}
```

sheet.go
```diff
@@ -103,6 +115,66 @@ func (s *Sheet) SetColWidth(startcol, endcol int, width float64) error {
...
+	for r, row := range s.Rows {
+		for c, cell := range row.Cells {
+			if cell.HMerge > 0 || cell.VMerge > 0 {
+				coord := fmt.Sprintf("%s%d", numericToLetters(c), r+1)
+				merged[coord] = cell
+			}
+		}
+	}
+
+	// Now go through the map and create the ghost cells
```
mafredri: Maybe a bit misleading? Ghost cells aren't actually created here, they are implicitly allocated by `s.Cell(x, y)`.

sheet.go
```diff
@@ -110,9 +182,15 @@ func (s *Sheet) makeXLSXSheet(refTable *RefTable, styles *xlsxStyleSheet) *xlsxW
 	maxRow := 0
 	maxCell := 0
 
+	// Scan through the sheet and see if there are any merged cells. If there
+	// are, we may need to extend the size of the sheet. There needs to be
+	// phantom cells underlying the area covered by the merged cell
```
mafredri: Same here re: phantom cells, extending sheet size isn't actually done here (it's implicit throughout the library now).

style.go
```diff
@@ -183,7 +183,8 @@ func DefaultFill() *Fill {
 }
 
 func DefaultBorder() *Border {
-	return NewBorder("none", "none", "none", "none")
+	//return NewBorder("none", "none", "none", "none")
+	return NewBorder("", "", "", "")
```
mafredri: Please remove the commented code, also, what's the purpose of not using `"none"` anymore?

xmlStyle_test.go
```diff
@@ -60,6 +60,8 @@ func (x *XMLStyleSuite) TestMarshalXlsxStyleSheetWithAFill(c *C) {
 }
 
 // Test we produce valid output for a style file with one border definition.
+// Test has been modified, because of Excel quirks, you need to provide empty
+// elements.
```
mafredri: I don't think it's useful to say what this function used to do. It should always only be a comment about what this function does. e.g.

```golang
// Test we produce valid output for a style file with one border definition.
// Empty elements are required to accommodate for Excel quirks.
```

^ I'm not exactly sure what it means, but along those lines, you as the author probably know better what to write :).

xmlStyle_test.go
```diff
@@ -69,9 +71,9 @@ func (x *XMLStyleSuite) TestMarshalXlsxStyleSheetWithABorder(c *C) {
 	border.Left.Style = "solid"
 	border.Top.Style = "none"
 	styles.Borders.Border[0] = border
-
 	expected := `<?xml version="1.0" encoding="UTF-8"?>
-<styleSheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"><borders count="1"><border><left style="solid"></left><top style="none"></top></border></borders></styleSheet>`
+<styleSheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"><borders count="1"><border><left style="solid"></left><right style=""></right><top style="none"></top><bottom style=""></bottom></border></borders></styleSheet>`
```
mafredri: What's the purpose of mixed `""` and `"none"` here, couldn't they all be `"none"`?

xmlStyle.go
```diff
@@ -687,56 +693,50 @@ func (borders *xlsxBorders) Marshal(outputBorderMap map[int]int) (result string,
...
 type xlsxBorder struct {
-	Left   xlsxLine `xml:"left,omitempty"`
-	Right  xlsxLine `xml:"right,omitempty"`
-	Top    xlsxLine `xml:"top,omitempty"`
-	Bottom xlsxLine `xml:"bottom,omitempty"`
+	Left     xlsxLine `xml:"left,omitempty"`
+	Right    xlsxLine `xml:"right,omitempty"`
+	Top      xlsxLine `xml:"top,omitempty"`
+	Bottom   xlsxLine `xml:"bottom,omitempty"`
+	Diagonal xlsxLine `xml:"diagonal,omitempty"`
```
mafredri: Diagonal doesn't seem to be supported anywhere else, is there any point in adding it here?

mafredri:
I finally took the time to properly review this PR, and I believe it's properly managed to address my concerns! I've done some simple testing and tested it out in a larger application that generates xlsx files with many borders and it seems to be doing the right thing there as well.

There's still some points that concern me:

1. This PR introduces a breaking change as demonstrated in [this gist](https://gist.github.com/mafredri/80d76e7672e84d33be8c). Previous behavior would create a full border around the merged cells. This PR which simplifies border logic breaks this old behavior and removes the right border of the merged cells.
2. The implicit allocation of cells caused by the change to `Sheet.Cell` might cause users to end up allocating a lot more memory (rows / cells) than they realize. Amending this is not relevant for this PR though, it would require rethinking how rows, cols and cells are represented in the library. Optimally unused cells would not take up any memory.

mafredri:
PS. I'd recommend the commits be squashed before merging.

style.go
greenenergy: Excel unfortunately isn't terribly robust, and files that work perfectly well with LibreOffice or Google Sheets may not look quite right. I was having trouble getting the borders to appear, so I was trying to bring the output of this library more in line with what Excel provides, in the hopes that it would be happier. And one of the differences is that Excel doesn't use "none", it uses empty quotes.

It turns out that this wasn't the problem with this particular function -- the problem turns out to be that Excel is expecting a set of borders to exist that are never used - it wants a set of empty borders as the first set, even if no style ever references them.

xmlStyle.go
greenenergy: Well, diagonal is part of the spec. Adding it here allows you to use it.

xmlStyle_test.go
greenenergy: There is no purpose to mixing -- it's a test that I missed I guess. They could all be "" or "none", and they may as well be "" for brevity and simplicity. Plus, it's what excel does.

sheet.go
greenenergy: I like your method better, I'll change it.

sheet.go
greenenergy: Well, this section's purpose in general is to create the ghost cells necessary for the merged areas. What do you think would be better phrasing?

sheet.go
greenenergy: I guess you could say that these are more "why" comments than "what" comments. Do you think that they are unhelpful or perhaps confusing?

greenenergy:
Regarding your gist - I think it's an invalid comparison, since the purpose of how the bordering works with merged cells is that it carries the borders over into the ghost cells. How else would it work? Your example  appears to expect a contradictory set of commands -- that at the top level, you don't want a right border, but in one of the ghost cells you do, so which wins? If we just stick to the idea that the border style applied to the first cell in the merge group gets reflected across all the ghost cells underneath, there is no issue.

sheet.go
mafredri: I suck at comments (trying to become better), but maybe something along the lines of:

```golang
// This loop iterates over all cells that should be merged and applied the correct
// borders to them depending on their position. If any cells required by the merge
// are missing, they will be allocated by s.Cell().
```

?

sheet.go
mafredri: I think a simple comment like `// Make sure cell border settings are adjusted before filling the new worksheet`. The comment above `handleMerged` should already be sufficient in describing what is happening.

style.go
mafredri: Thanks for the explanation, however I'm not quite sure what the takeaway here was, can we revert this change or should we remove the commented code? :D

greenenergy: I have removed the commented code.

xmlStyle.go
mafredri: How can a user ever apply it since it's not part of the `Border` struct in `styles.go`? I'd also think it would need to be supported by the `xlsxBorder.Marshal` function.

sheet.go
greenenergy: Except that I'm not changing the border settings, I'm literally accessing those phantom cell locations to create the underlying cells. Do you feel that the comment is confusing, or just superfluous?

sheet.go
greenenergy: Ok

greenenergy:
As for squashing - there have been a number of merges while this side branch has been in development. I'm not exactly a git master, and squashing commits that encompass merges seems to be a nontrivial operation, fraught with danger. :) Is there a reason beyond aesthetics that this must be done?

mafredri:
@greenenergy I disagree with the invalidity of my comparison. My statement was that this PR introduces a breaking change in behavior (as compared to current master). What used to work no longer doesn't, to me that's a breaking change. You could even argue that the way this previously worked, the expected way of working with merged cell borders is exactly as in my gist (it was for me). You could of course apply ALL borders to all cells and end up with the correct borders, but when you set a border and it has no effect, why set it at all, instead micromanage the borders for all cells.

To further prove that this breaks old behavior, here's a comparison of what was possible (left) and what happens with this PR (right):

xmlStyle.go
greenenergy: Ok, removing it.

mafredri:
Regarding squashing, I'd say aesthetics is definitely one, but another is removing noise and making it easier to track the reason for a set of changes (as in this PR for example). Either way, I'm not the owner of this repo so this is not up to me, feel free to do as you please with it :).

greenenergy:
Ok, fair enough. It is a change in behaviour. I guess the question is - is it an improvement? Which method is more useful and/or more obvious to the users? I would argue that this method is; and it's a lot easier to explain to my users how to create a bordered merge box this way than having to figure out where the phantom cells are under the merge cells and manually border them.

sheet.go
mafredri: I guess we think about this differently. To me the "creating cells" is just a side-effect. They might very well already be created by the user, for example. To me `handleMerged` isn't a cell-creating function, it's a border adjusting one. It makes sure that all cells included by the merge has the correct border.

But to actually answer your quesiton, maybe I think it's slightly superfluous in the fact that it's describing (partially) the behavior of another function, e.g. `handleMerged`. But reading it again I think mentioning the border adjusting there would be sufficient.

I don't feel strongly about this though, you can leave it as-is if you want.

greenenergy:
And regarding squashing - I see that git here reports 16 commits on this branch - but when I try to do a rebase back to the beginning of this change, it incorporates all the changes from everyone else who's merged in, and I really don't want to do that.

mafredri:
> I guess the question is - is it an improvement?

I'd argue yes, one gripe I've had is definitely with adding borders. This should take most of the pain away and please most users. Documenting it would definitely be a bonus though. Even something simple as "Merged cells lose their style".

greenenergy:
Ok - so at this point, are you ok with the changes, or do you still feel that something else needs to be addressed?

mafredri:
I'm definitely OK with it. Final say is with @ShawnMilo and @tealeg though.

ShawnMilo:
For the record, it's @tealeg's project, although I am a committer. I'm not well-versed in styling issues, but the PR isn't that big, so I'll take a look.

As for rebasing, there's an easy fix; just diff your branch against the current `master` and export that as a patch. Then check out a fresh copy of `master` and apply your patch.

```
git diff --no-prefix master > borders.patch
git checkout master
patch -p0 < borders.patch
git commit -a
```

That'll give you a one-commit branch, and you can make a PR on that. Then, if you want, you can push that branch to this PR:

```
git push -f origin/mybranch this_patched_branch
```

sheet.go
```diff
@@ -78,10 +78,17 @@ func (s *Sheet) Col(idx int) *Col {
 // containing the data from the field "A1" on the spreadsheet.
 func (sh *Sheet) Cell(row, col int) *Cell {
 
-	if len(sh.Rows) > row && sh.Rows[row] != nil && len(sh.Rows[row].Cells) > col {
-		return sh.Rows[row].Cells[col]
+	// If the user requests a row beyond what we have, then extend.
+	for len(sh.Rows) <= row {
```
ShawnMilo: Since xlsx files don't contain data for non-existent rows or columns, it seems like this would unnecessarily bloat the resulting file with potentially hundreds of thousands of empty rows. Have you manually tested this using, for example, row 100,000?

greenenergy: This code was put in place because when you merge cells, ghost cells have to be generated underneath the area covered by the merge.

As far as I understand, the library is not set up to deal with a sparse matrix, so there is no mechanism to create a cell at, say, 100,000x100,000 and not backfill all the other cells, since cells are addressed via array indexes.

So yes, a consequence of this is that if someone puts a cell down at 100,000 you're going to have a big sheet generated. The way it was before, it would just crash because it would try to address way beyond the end of the slice.

Unless, of course, I am missing something regarding the core of how rows and cells are accessed.

greenenergy:
Thanks @ShawnMilo re the rebasing tip. I'll take care of that.

greenenergy:
@ShawnMilo Do you have any further concerns with the PR? If not, would you mind doing the merge? It looks like there is a bug in the library (something unrelated to this code) that we'd like to work on, but don't want to try to manage multiple simultaneous side branches.

ShawnMilo:
Honestly, I've never really messed with styling in spreadsheets, so even though I've looked it over I don't feel qualified to merge it without @tealeg giving the go-ahead. I don't want to cause some subtle problem.

I completely sympathize with the multiple-branch pain. Sorry I can't solve it. I've noticed that Geoff hasn't participated in this thread; perhaps he's on vacation or otherwise occupied. I know it's not a great solution, but you can always work off of your fork temporarily.

greenenergy:
Ok, thanks for the update. :)

tealeg:
This is a large, and useful patch.  Though there may be some knock-on impact of this I'm inclined to agree it's the right thing and deal with bugs as they come up.

Apologies for not being more involved, but well done to @greenenergy, @mafredri, @ShawnMilo for your diligence here!

PR #150, opened 2015-07-23
lightGrey should actually be lightGray or gray125

mafredri:
While investigating issues with Gmail not rendering xlsx preview I noticed that one problem was that greyFill was incorrectly set to `lightGrey` instead of `lightGray`.


tealeg:
@mafredri good spot!  I'm afraid that's the sort of mistake I make a lot ;-)

PR #155, opened 2015-08-07
added WrapText to alignment

rbucker:
The code is implemented, however, it is not actually causing the cell to wrap. This may only be part of the solution. If there is more or if wraptext is not possible then this PR can be dropped.


tealeg:
@rbucker it would be a good idea to add tests to be sure that this value is making it through to the file that is written.

tealeg:
I went ahead and did some minimal testing and changes to ensure the data makes it to the file.  Be sure to set the Style.ApplyAlignment to true if you set values in the alignment.

PR #156, opened 2015-08-12
Change xml exporting shapes

blackss2:
prevent all tabSelected be true
prevent row.cell.V showing up when Value is null
change default celltype to CellTypeGeneral from CellTypeString
prevent sharedString.xml file creation when no string shared
add emptyomit to unnecessary attributes


blackss2:
clone for merging(update my code)

PR #157, opened 2015-08-12
tablSelected, empty value, SAS import

blackss2:
fix multiple tabSelected problem by adding sheet.Selected(default sheet[0] : true, else : false)
omitempty cell.V node when CellTypeString and value length is zero
fix workbook attribute for SAS import(if you set column type as Date and cell.SetDate(t), SAS will recognize it as Date!)


tealeg:
Hi @blackss2 

Urgh!  I hate that we have to do those namespace hacks.  I've pushed the existing Numbers related hacks into the same function, renamed it and changed the comment.  Otherwise the rest is merged as is.  Thanks for your efforts.

PR #161, opened 2015-09-11
export the cell numeric format so a formula cell can be set to percent

mattetti:
…for instance.

This is a very straight forward PR, nothing fancy. I found myself wanting to format my cells many different ways (various currencies, percentage etc..). I know that in most cases I could have used `SetFloatWithFormat` but I needed to apply the format to the result of a formula. I could have added a `SetFormulaWithFormat` but even with that, I think that it makes a lot of sense to export the format field. After all, this is something Excel users can access directly and there are a lot of edge cases where it might be needed.

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -30,7 +30,7 @@ type Cell struct {
 	Value    string
 	formula  string
 	style    *Style
-	numFmt   string
+	NumFmt   string
 	date1904 bool
 	Hidden   bool
 	HMerge   int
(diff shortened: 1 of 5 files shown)


ShawnMilo:
I don't have any technical objections. All that was done was `numFmt` was replaced with `NumFmt` in a bunch of places.

@tealeg: Do you have any philosophical problems with exposing this field?

tealeg:
No objection, it won't break things, so let's try it.

PR #162, opened 2015-09-18
Set ApplyNumberFormat in handleNumFmtIdForXLSX

### LINKED ISSUE #158: Number format strings do not work in LibreOffice, but do in Excel
```
package main

import (
    "fmt"
    "github.com/tealeg/xlsx"
)

func main() {
    var file *xlsx.File
    var sheet *xlsx.Sheet
    var row *xlsx.Row
    var cell *xlsx.Cell
    var err error

    file = xlsx.NewFile()
    sheet = file.AddSheet("Sheet1")
    row = sheet.AddRow()
    cell = row.AddCell()
    cell.SetFloatWithFormat(42000, "0.00e+00")
    err = file.Save("MyXLSXFile.xlsx")
    if err != nil {
        fmt.Printf(err.Error())
    }
}
```
creates a file which shows "4.20E+00" in Excel 2013 but "42000" in LibreOffice 4.4.2.2

pwaller:
As mentioned in #158, applyNumberFormat was always 0 in the output.

For some reason, Excel didn't care, but all the other clients I tried (Google docs, OpenOffice, Gnumeric) did care.

Fixes #158.

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -277,6 +277,9 @@ func handleStyleForXLSX(style *Style, NumFmtId int, styles *xlsxStyleSheet) (XfI
 func handleNumFmtIdForXLSX(NumFmtId int, styles *xlsxStyleSheet) (XfId int) {
 	xCellXf := makeXLSXCellElement()
 	xCellXf.NumFmtId = NumFmtId
+	if xCellXf.NumFmtId > 0 {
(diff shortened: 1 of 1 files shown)


tealeg:
Thanks very much!

Issue #147, opened 2015-07-16
sheet.Rows does sometimes contain nil Rows


claudiofelber:
There are XLSX-Files that result in sheet.Rows containing single nil entries. The behavior can be reproduced by loading the following sample file: http://ge.tt/88p9mTK2/v/0

It is probably due to `makeEmptyRow()` not being called for each row that has no definition in the XLSX-File. For example the following simple modification in `readRowsFromSheet()` fixes the problem (although not being very efficient):

``` go
// insert leading empty rows that is in front of minRow
for rowIndex := 0; rowIndex < maxRow; rowIndex++ {
    rows[rowIndex] = makeEmptyRow()
}
```

**Note:** `minRow` has been replaced with `maxRow`in the loop condition, thus ensuring that all rows will be properly initialized.

ShawnMilo:
@claudiofelber: Is this resolved in `master`? It looks like #168 may have fixed it.

PR #167, opened 2015-09-26
WIP: 147 nil row fix

ShawnMilo:
This is a work-in-progress because there are no tests.

@claudiofelber: The solution you mentioned in #147 was good, but checking on `rowCount` is better than  `maxRow`, because that would ignore the final row.

@tealeg: I had a lot of trouble writing a test for this. I actually made this fix weeks ago, but never pushed it due to that difficulty. I tried copy/pasting an existing test and changing the XML to put skip row numbers, but I couldn't get the desired output. Do you have any suggestions?

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"path"
+    "log"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


ShawnMilo:
Closing, since it's now a duplicate of #168.

PR #168, opened 2015-09-28
Fix "off by one" bug when the spreadsheet omits empty rows

boxofrad:
This fixes an issue where a row would be accidentally overwritten with a blank row, if the following (blank) row was omitted in the worksheet file.

[Here is a file](https://s3-eu-west-1.amazonaws.com/boxofrad/Shared/bug.xlsx) (exported from Google Sheets) that exposes the problem - It's not possible to access the cell at reference: **A4**.

I'm not sure what the original reason for subtracting the `minRow` was, but it seems erroneous to me.

Thanks! :sparkling_heart:

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -487,9 +487,8 @@ func readRowsFromSheet(Worksheet *xlsxWorksheet, file *File, sheet *Sheet) ([]*R
 		// stored data
 		for rawrow.R > (insertRowIndex + 1) {
 			// Put an empty Row into the array
-			index := insertRowIndex - minRow
-			if index < numRows {
-				rows[index] = makeEmptyRow(sheet)
+			if insertRowIndex < numRows {
+				rows[insertRowIndex] = makeEmptyRow(sheet)
 			}
 			insertRowIndex++
 		}
--- a/lib_test.go
+++ b/lib_test.go
@@ -1068,3 +1068,60 @@ func (l *LibSuite) TestFormulaForCellPanic(c *C) {
 	// panic is the real win.
 	c.Assert(formulaForCell(cell, sharedFormulas), Equals, "")
 }
+
+func (l *LibSuite) TestRowNotOverwrittenWhenFollowedByEmptyRow(c *C) {
+	sheetXML := bytes.NewBufferString(`
+	<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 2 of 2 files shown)


ShawnMilo:
Thanks for the PR with a test!

This is an exact duplicate of #147, and your fix is very similar to my PR, #167. However, since your PR has a test and mine does not, I would lean towards closing mine and accepting yours.

One thing -- instead of `numRows`, I think `rowCount` should be used, since `numRows` would stop one short. Please check that out and see if you agree, or whether the nature of the problem means that you don't need to care if it's the last row.

boxofrad:
Thanks for the quick response @ShawnMilo, and for pointing me towards those other issues... I hadn't seen them! :smile: 

I'm not too familiar with this code (so couldn't confidently comment on `numRows`) but I think you're right (that it'd stop one short).

boxofrad:
I've just double checked and I think that `numRows` is actually synonymous with `rowCount` because `rowCount` is the length given to `make` when creating the slice...

We can probably just swap them out if you like?

ShawnMilo:
Yeah, I think you're right, and that I wasn't thinking about it properly. They should be equal.

```go
numRows := len(rows)
rowCount = maxRow + 1
```

It actually looks like `numRows` can be removed and `rowCount` used there as well. @tealeg, does that sound right to you? Tests pass with this patch:

```diff
                rows[rowIndex] = makeEmptyRow(sheet)
        }
 
-       numRows := len(rows)
        for rowIndex := 0; rowIndex < len(Worksheet.SheetData.Row); rowIndex++ {
                rawrow := Worksheet.SheetData.Row[rowIndex]
                // Some spreadsheets will omit blank rows from the
@@ -488,7 +487,7 @@ func readRowsFromSheet(Worksheet *xlsxWorksheet, file *File, sheet *Sheet) ([]*R
                for rawrow.R > (insertRowIndex + 1) {
                        // Put an empty Row into the array
                        index := insertRowIndex - minRow
-                       if index < numRows {
+                       if index < rowCount {
                                rows[index] = makeEmptyRow(sheet)
                        }
                        insertRowIndex++
```

tealeg:
@ShawnMilo that patch looks ok.  I'm half remembering there being a issue with files with leading blank rows, but if there aren't tests for it, then it's not yet a maintainable use case.  Better to make this change and expose the problem again (then add tests!).

PR #169, opened 2015-09-30
add OpenBinary & OpenReaderAt

blackss2:
reading xlsx file from raw bytes(useful for web server)


tealeg:
Thanks!

firewaternottogther:
OpenBinary parses just 5M XLSX files, too slow。

PR #171, opened 2015-10-02
WriteSlice supports writing []interface{}

klniu:
Now WriteSlice can write slice with the type []interface{}, for example:
```go
s := []interface{}{"test", 1, true}
row.WriteSlice(&s,-1)
```


tealeg:
@klniu awesome work!

PR #176, opened 2015-10-23
Fix sample code on README

snoozer05:
receive a second return value

ref: https://godoc.org/github.com/tealeg/xlsx#File.AddSheet


tealeg:
Good catch @snoozer05 !

PR #177, opened 2015-11-02
add write file callback

sorawa:
add write file callback


tealeg:
Hi @sorawa.  This is a nice idea.  Could you please add a new method called SaveWithCallback (or similar) rather than changing the contract of Save.  Also, please 1. make sure all existing tests are passing and 2. add tests for the new behaviour.

tealeg:
I'm assuming this isn't going to happen.

PR #181, opened 2015-11-06
fmt.Stringer interface support in WriteSlice & WriteStruct

ivnivnch:
[fmt.Stringer](https://golang.org/pkg/fmt/#Stringer) support for struct fields in WriteStruct and for slice values in WriteSlice


tealeg:
@ivnivnch I like it!

PR #196, opened 2016-03-04
Skip string literals when trying to shift cells in shared formula

thieupham-wf:
### Problem / Feature

Will skip string literals in a formula, so we don't unintentionally shift those looking like cell labels.

For example, ```IF(C23>=E$12,"Q4",IF(C23>=$D$12,"Q3",IF(C23>=C$12,"Q2","Q1")))```, "Q4", "Q3", "Q2", and "Q1" should not be shifted.

Please review @tealeg  
Thanks!


tealeg:
@thieupham-wf thanks for your work!

thieupham-wf:
@tealeg Thanks!

Issue #191, opened 2016-01-14
Empty cells with formatting fails to parse


eproxus:
A empty cell with Number format and a background color assigned fails to parse with the following error:

    strconv.ParseFloat: parsing "": invalid syntax

ShawnMilo:
Please supply a simple spreadsheet that duplicates the problem. That will make it easy to replicate, and then fix.

eproxus:
Here's a workbook that reproduces the error:
[Workbook1.xlsx](https://github.com/tealeg/xlsx/files/90723/Workbook1.xlsx)

ShawnMilo:
Thanks for the file. I was able to open the file without an error. Sorry to ask again (I should have asked before), but please provide a simple, *complete* Go application which reproduces the error.

eproxus:
Hi again, here's the program I used:

```go
package main

import (
	"fmt"

	"github.com/tealeg/xlsx"
)

func main() {
	excelFileName := "/path/to/Workbook1.xlsx"
	xlFile, err := xlsx.OpenFile(excelFileName)
	if err != nil {
		fmt.Print("Error")
	}
	for sheetN, sheet := range xlFile.Sheets {
		for rowN, row := range sheet.Rows {
			for cellN, cell := range row.Cells {
				fmt.Printf("%d %d %d %s\n", sheetN, rowN, cellN, cell.String())
			}
		}
	}
}
```

eproxus:
Just realized that maybe `cell.String()` is causing the problem. Should I use something else to access number contents (and if so, how do I know when to)?

ShawnMilo:
Yeah, if you use `cell.Value` instead of `cell.String()` it works.

First, the cell has a number format (click on the cell, then go to Format->Cell), and you'll see it is supposed to be a number. However, the value in the cell is "" (empty string). This causes a formatting error. If the cell was a string cell, it should be fine.

Secondly, (and this is an issue for the `xlsx` maintainers to discuss), the `FormattedValue` function returns a string, and no error. if it receives an error, then it passes the `err.Error()` (a string) back, and that's what you get in your code. Your code doesn't break, but you get Go error messages in your data.

[...]

PR #198, opened 2016-03-06
Fix for 191 formatted value return error.

ShawnMilo:
As discussed in the comments in #191, fixes FormattedValue to return `(string, error)` instead of just `string`.

Previously, if there was an error, the `error.Error()` was just returned as the string value. This worked, but resulted in error-free execution with the potential for bad output.

See 9e8fb11ea051b80749b9a18fdcbc97ebca291b87 for the actual, very minor, change.

Then commit c88ad73d9b7d4d9c8c60b4726a67d5a6e602cf4d for the test changes that had to be done.

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -68,7 +68,7 @@ func (c *Cell) SetString(s string) {
 }
 
 // String returns the value of a Cell as a string.
-func (c *Cell) String() string {
+func (c *Cell) String() (string, error) {
 	return c.FormattedValue()
 }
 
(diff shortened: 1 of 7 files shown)


ShawnMilo:
A third commit has been added to make the tests less verbose.

tealeg:
Wow, @ShawnMilo  that's a chunk of work!  Tests out and I couldn't see anything wrong in there, or missed.  

Thanks a lot for that.

ShawnMilo:
You're welcome. I wrote scripts to generate most of the changes, of course. ^_^

Issue #192, opened 2016-01-14
Open and Save xlsx Error


qhsong:
I use Mac Office 2013 to generator a xlsx,and Use following code to edit xlsx file and open `test2.xlsx`, it tell me this file need repair.
```go

func main() {
	file, err := xlsx.OpenFile("test.xlsx")
	if err != nil {
		log.Fatal(err)
	}

	sheet := file.Sheets[0]
	row := sheet.AddRow()
	row.AddCell().SetValue("12313131")
	err = file.Save("test2.xlsx")
	if err != nil {
		log.Fatal(err)
	}
}

```

tealeg:
We have experienced several issues of this kind.  If you could provide a sample of the file before and after repair that would be very useful.  I personally don't have access to Mac Office 2013.

shawnye:
I upload an attatchment , just for test , 
The test.xlsx contains no formula , the test_error.xlsx does. The corresponding counterpart show the result of `saveas` action, as described in main.go.
test_saveas.xlsx (***The style is NOT completely copied***)
test_error_saveas.xlsx (***Cant not open, even repair***)


Built on go version go1.6 windows/386
[test_saveas.zip](https://github.com/tealeg/xlsx/files/145608/test_saveas.zip)

qhsong:
Sorry for delay, I will upload this file later.

qhsong:
[Uploading test2.xlsx…]()
[test2_After_repair.xlsx](https://github.com/tealeg/xlsx/files/148100/test2_After_repair.xlsx)

test2.xlsx is not repair file and test2_After_repair.xlsx have been repaired by Mac Office 2013.

TaylorBarrick:
The files attached above are also causing the same issue in Windows with Excel 2010.  When comparing the files, it seems that, while there are several differences, the only blocking file is the styles.xml.

Before Repair:
```
<styleSheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
     <fonts count="1">
          <font>
               <sz val="12" />
               <name val="DengXian" />
               <family val="2" />
               <charset val="134" />
               <color rgb="FF000000" />
          </font>
     </fonts>
     <fills count="2">
          <fill>
               <patternFill patternType="none" />
          </fill>
          <fill>
               <patternFill patternType="lightGray" />
          </fill>
     </fills>
     <cellStyleXfs count="1">
[...]

Issue #151, opened 2015-07-23
Incorrect use of cellStyleXfs and cellXfs


mafredri:
While investigating why Gmail refuses to render an xlsx document generated by this lib, I noticed two issues. The first on is taken care of by PR #150, but the second one is a bit tougher (I have not yet looked at the relevant code):

As I understand the spec, [cellStyleXfs](https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.cellstyleformats(v=office.14).aspx) should contain "master"-formats for cells, and [cellXfs](https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.cellformats(v=office.14).aspx) should contain formats for the different kinds of cells that reference a "master" format. I think the normal situation is that there is **1** element in `cellStyleXfs` and all `cellXfs` reference it through `xfId` (which is missing btw!).

I have pasted the differences between a document generated by tealeg/xlsx and then saved with Excel in [this gist](https://gist.github.com/mafredri/6cf6920a23a6a60ec536).

What are the thoughts on this? I'm unsure if it's sufficient to create one generic hard-coded `cellStyleXfs` or if there are some cases we're not considering by doing this.

mafredri:
Ok, so no issue is complete without providing a test case as well. I tried to create a minimal sample of the type of document I'm creating. I've included the xlsx files, extracted files (pretty-printed xml), a diff (comparison.diff) of the pretty-printed xml files.

I noticed that if I only copy over `style.xml` from the Excel "fixed" file, the columns are messed up. So relevant is BOTH `style.xml` and `sheet1.xml`. Copying over both of these allows Gmail to render the xlsx same as if it was saved by Excel.

And finally, but not least: a screenshot of how it looks in Gmail:
![gmail-comparison](https://cloud.githubusercontent.com/assets/147409/8851061/9caea664-3155-11e5-9da6-4f6e1eb0afc9.png)

tealeg:
@mafredri  I've investigated this issue a little in the past.  It does need addressing.  I'd be happy to see patches as I don't really have the time right now.

mafredri:
@tealeg alright, at least we have it documented now. I'll see if I have some extra time to poke around on this at some point.

pwaller:
I'm taking a look at this in the context of #158.

@mafredri 

> I have pasted the differences between a document generated by tealeg/xlsx and then saved with Excel in this gist.

[...]

Issue #133, opened 2015-07-02
Opening an xlsx file and saving it as another xlsx file creates an invalid file


manuviswam:
``` go
package main

import (
    "fmt"
    "github.com/tealeg/xlsx"
)

func main() {
    var file *xlsx.File
    var err error

    file, err = xlsx.OpenFile("/path/to/file.xlsx")
    err = file.Save("/path/to/anotherFile.xlsx")
    if err != nil {
        fmt.Printf(err.Error())
    }
}
```

The above code compiles and runs fine. It creates anotherFile.xlsx but when opening this file Excel application crashes.

Immortalin:
@tealeg I also faces this issue. However in my case office 2010 was able to "repair" the file. My suggestion is to add a file.SaveOpenFile method and utilize (this)[https://golang.org/pkg/os/#File.Close] to close a file after writing to it.

ShawnMilo:
I am not able to duplicate this with LibreOffice.

@manuviswam: Are there any error messages from Excel?

@Immortalin: Can you provide an example of a very simple .xlsx file before and after the repair process, so I can see what the difference is?

tealeg:
@ShawnMilo you can almost never reproduce such errors with LibreOffice - it has a much more liberal parsers.  In most cases these issues stem from small variances in what is output from what the XSD defines (noting, for example that element ordering is important for Excel to consider a document valid). 

For what it's worth - eliminating this kind of problem is one of the reasons I've started work on some XSD work.  As Shawn is doing an excellent job here, I'm trying to use my spare cycles to progress that work and fix the big problem rather than worrying about all the small ones ;-)

@Immortalin I'm not really sure I understand your suggestion.  Did you think that the file was missing some data that hadn't been flushed to it?  There isn't a general problem with the validity of files, per se, it is only that Excel has an extremely picky validation process and is actually terrible at telling you what the real error is.   The before and after version of the files are the best way we currently have of seeing what's really wrong.

tealeg:
Note also that this is essentially a duplicate of #53 - however that is only because Excel doesn't given us sensible error messages in these cases.

manuviswam:
@ShawnMilo I am also able to repair it with the latest version of MS Office
Please find the sample files below
[Before repair](http://wikisend.com/download/289624/Output.xlsx)
[After Repair](http://wikisend.com/download/113338/OutputAfterRepair.xlsx)

Patrick-H:
[...]

Issue #84, opened 2015-01-29
XLSX Style doesn't work


forhappy:
Hi, I'm using the  XLSX style, e.g Fill style in a Cell, I set cell.SetStyle(style) where style is fill := NewFill("solid", "#12CDD3", "#FFFFFF"); style.Fill = *fill,  but it does not show any fill color in the generated XLSX file, Am I using the style the right way, what I was missing here? I set ApplyFill to true by the way. ;-) Thanks

tealeg:
Hi @forhappy  - the style support is currently not working, as we've had trouble with corrupted file output.  I plan to re-enable it as soon as I have some spare time.  I am preparing for the arrival of my 2nd child right now, for obvious reasons that is slowing down progress.

EricLagerg:
@tealeg if this is still an issue I could look into it. (Also, congrats on the child.)

tealeg:
@EricLagerg it's almost definitely still an issue.  Long term what I think is needed is to generate the underlying structs from the XML Schema and introduce test that validate output aginst that schema.  Excel in partitular is very stirct about compliance - it epect elments to be ordered as per the schema, for example.

Still waiting on our 2nd daughter, should be any day now.  Thank for the wishes!

EricLagerg:
@tealeg schema.microsoft.com and the other websites aren't loading for me. Do you have any schema resources available?

EricLagerg:
@tealeg I also noticed in some .xlsx files (and in Msoft's C# docs) there are some styling aspects that this project skips over. One example would be `<bgColor indexed=... />`, found here: https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.indexedcolors(v=office.14).aspx

What are your thoughts on changing the high-level `Fill` struct to one something like this:

`type Fill struct {
        PatternType string    // or an enum value..?
        BgColor        BgType // obviously a better name
        FgColor        FgType  // same as above
}

type BgType struct {
        ColorModel string // indexed, rgb, etc.
        Color           string // "64", "FF8DB3E2", etc.
}`

Or just adding more members to `Fill`?

tealeg:
@EricLagerg I'm happy to see more functionality in the highest level.  If we're going to remove anything, or move anything that's visible to the end user of the library then we need good justification, and possibly to branch off a new development version of the project (something I've been considering).  

[...]

PR #206, opened 2016-03-16
Fix styles

tealeg:
This branch should drastically improve the styles files output by XLSX, and also round-trips thereof.

Up until now we've always incorrectly handled the cellStyles, cellStyle and cellStyleXfs (and there child Xfs).    The original (wrong) assumption was that there was a cellStyleXfs.Xf that corresponded to every cellXfs.Xf and that the worksheet cells "s" attribute was an index to both.   In reality the worksheet cell references only the cellXf.Xf element.  That Xf element may optionally set an XfId of it's own thata references a cellStyleXfs.Xf (which can also be referenced by Name from a cellStyle).

I make no attempt here to do anything with named styles (cellStyles) in the interface of this library. I only marshal and unmarshal them correctly, and link them to via the Xf.XfId correctly.  

This should address (at least partially):
- #192 
- #151 
- #133 
- #84 

@ShawnMilo please could you at least sanity check the code ;-)


ShawnMilo:
I looked it over and put out requests to those suffering from those bugs to do a local test. Hopefully that will flush out any edge-case bugs in addition to confirming (or not) the fix.

I didn't see anything that jumped out at me, but I'm going to look again, and in the meantime I hope we get some responses from those who were participating in those issues.

tealeg:
Cheers @ShawnMilo on both counts.  I'm hoping this fixes these repair issues - or at least some of them.   For what it's worth I think every attribute needs to be converted to a pointer to a value eventually so we can really tell the difference between attribute="" and the attribute being omitted.

PR #207, opened 2016-03-23
Fixed DefinedNames.DefinedName pointer issue.

Joshuabaker2:
file.DefinedNames would be an array of the proper size, but all values would be the last &entry, as they were all referencing the same memory location. 

e.g. 
**Memory addresses and values before fix (file with three defined names):**
[0xc8201363c0 0xc8201363c0 0xc8201363c0]

&{Sheet1!$E$5:$G$7 NamedRange3       0 0 false false false false false false}
&{Sheet1!$E$5:$G$7 NamedRange3       0 0 false false false false false false}
&{Sheet1!$E$5:$G$7 NamedRange3       0 0 false false false false false false}

**Memory addresses and values after fix:**
[0xc820136280 0xc820136318 0xc8201363b0]

&{Sheet1!$A$1:$C$2 NamedRange1       0 0 false false false false false false}
&{Sheet1!$E$1:$G$2 NamedRange2       0 0 false false false false false false}
&{Sheet1!$E$5:$G$7 NamedRange3       0 0 false false false false false false}

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -356,14 +356,23 @@ func parseTime(c *Cell) (string, error) {
 	}
 	val := TimeFromExcelTime(f, c.date1904)
 	format := c.GetNumberFormat()
+
+
 	// Replace Excel placeholders with Go time placeholders.
 	// For example, replace yyyy with 2006. These are in a specific order,
 	// due to the fact that m is used in month, minute, and am/pm. It would
 	// be easier to fix that with regular expressions, but if it's possible
 	// to keep this simple it would be easier to maintain.
+	// Full-length month and days (e.g. March, Tuesday) have letters in them that would be replaced
+	// by other characters below (such as the 'h' in March, or the 'd' in Tuesday) below.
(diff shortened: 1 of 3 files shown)


Joshuabaker2:
Continuing to use this library and found that there wasn't support for "mmmm" and "dddd" in the excel date type parser (e.g. "January" or "Monday"). Added support for these.

cell.go
```diff
@@ -377,6 +386,8 @@ func parseTime(c *Cell) (string, error) {
 		{"am/pm", "pm"},
 		{"m/", "1/"},
 		{".0", ".9999"},
+		{"%%%%", "January"},
+		{"&&&&", "Monday"},
```
tealeg: These changes require corresponding changes to the unit tests to ensure that they continue to work in the future.

Joshuabaker2: Good call. Unit tests added.

tealeg: Perfect thanks!

tealeg:
Thanks for this work @Joshuabaker2.

PR #216, opened 2016-05-17
stylistic changes

EricLagergren:
@tealeg: Love this library. 

I was using it again at work and noticed a fair amount of the code doesn't necessarily adhere to [typical Go conventions](https://github.com/golang/go/wiki/CodeReviewComments).

There's still a bunch more to clean up, but I figured I'd submit this to make sure you want it before I spend more time.

Most changes are things like returning literals, removing redundant declarations (e.g., `var x int = 0` which could just be `var x int`), and moving variables closer to their use (seemed very C-like to me).

### DIFF
--- a/xmlStyle.go
+++ b/xmlStyle.go
@@ -80,18 +80,18 @@ type xlsxStyleSheet struct {
 	CellXfs      xlsxCellXfs       `xml:"cellXfs,omitempty"`
 	NumFmts      xlsxNumFmts       `xml:"numFmts,omitempty"`
 
-	theme          *theme
+	theme *theme
+
+	sync.RWMutex   // protects the following
 	styleCache     map[int]*Style
 	numFmtRefTable map[int]xlsxNumFmt
-	lock           *sync.RWMutex
 }
 
 func newXlsxStyleSheet(t *theme) *xlsxStyleSheet {
-	stylesheet := new(xlsxStyleSheet)
-	stylesheet.theme = t
-	stylesheet.styleCache = make(map[int]*Style)
-	stylesheet.lock = new(sync.RWMutex)
-	return stylesheet
+	return &xlsxStyleSheet{
+		theme:      t,
+		styleCache: make(map[int]*Style),
+	}
 }
 
 func (styles *xlsxStyleSheet) reset() {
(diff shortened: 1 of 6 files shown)


EricLagergren:
All code passes `go test -v -race`, fwiw.

HeinOldewage:
@EricLagergren I like the changes you have made. I think they will add to overall readability.

@tealeg , @ShawnMilo :  I was wondering what path this library is taking towards correctness in general?
 
There are changes that can be made to the library to make it more robust but will slow it down.
(for example `getCoordsFromCellIDString` would parse "x1x1!#" as xx11, interpreting it as a valid cell position)
In other cases there are errors and race conditions that occur only when the file being read is broken.
(for example run `go test -race -tags=fuzzy -test.seed=1463474672`) 

Fixing `getCoordsFromCellIDString` will add a regex per cell in the file that will slow down the library without adding any benefit when parsing correct files, but we can detect broken files earlier.
The second error can be fixed by using channels in an idiomatic way, but will require changing a few functions (none of them public).

I mention these two cases only as examples to get a general feel for the rules we are trying to follow.
And on that note, maybe we should create a wiki, listing some of the frequently asked question we get in the issues and list some general guidelines for contributors.

Should I rather move this to an issue?

ShawnMilo:
@EricLagergren Thanks for the PR! I haven't finished reviewing it, but it looks great.

As for the project's attitude on correctness, I think I can safely speak for @tealeg when I say we're all for it!

I came on to the project late, but I suspect a lot of the code is really old (or copy/pasted from old code) and was just never cleaned up because this is a side project for both of us, and neither use it anymore for our day jobs. So when we have time it tends to go more towards fixing bugs.

Energetic contributors are always welcome!

xmlStyle.go
```diff
@@ -80,18 +80,18 @@ type xlsxStyleSheet struct {
 	CellXfs      xlsxCellXfs       `xml:"cellXfs,omitempty"`
 	NumFmts      xlsxNumFmts       `xml:"numFmts,omitempty"`
 
-	theme          *theme
+	theme *theme
+
+	sync.RWMutex   // protects the following
 	styleCache     map[int]*Style
 	numFmtRefTable map[int]xlsxNumFmt
-	lock           *sync.RWMutex
```
ShawnMilo: I'd argue for leaving the `lock` here, because `Alignment` is exported, meaning clients using the library could call methods on the `RWMutex`. Yes, they'd have to go out of their way to do that (or make a "typo"), but better safe than sorry.

ShawnMilo:
@EricLagergren Wow, looks great! Thanks for putting the time into this.

I left one tiny, almost nitpick note, but it's valid. Other than that I think it's ready to merge.

Sorry it took me so long to review. Day job and all that. The PR is definitely appreciated!

xmlStyle.go
EricLagergren: The lock is still there, just moved above what it protects. (Only places it was used internally was for the maps.)

EricLagergren: Oh! You mean unexport it? Sorry I misread your comment. Yes I fully agree, I will change that!

EricLagergren:
It's all good, I'm in the same boat. :)

If you'd like, I'll submit some more changes. I only made some just to make sure it was okay before I spent too much time on it.

xmlStyle.go
ShawnMilo: Yeah, that's all I meant -- keep it unexported. I could have been more clear. Thanks!

tealeg:
Just to say that @ShawnMilo is correct in everything he says.  This project is in a very real sense maintained by its community, and I'm extremely happy to see the amount of effort people put in to making it better.  For me "correct" trumps "fast", but I don't (for example) think that it's necessary for a library like this to deal with incorrect user input in any way other than giving a reasonable error (and not outright crashing the program it's used in).

At one point I had a bunch of plans for reworking this library, the only one that really bugs me is that there is a bunch of code that generates XML snippets "by hand" rather than using Go's XML marshalling.  This was a "qucik-fix' (or "hack" if you prefer) that I strongly regret.

PR #218, opened 2016-05-20
Add basic grouping support

xStrom:
This patch adds basic support for groups. Nothing fancy, you have to manually set the row/column levels. The only automatic thing right now is maximum level determining in the `makeXLSXSheet` method.


tealeg:
Awesome!  Thanks @xStrom

PR #219, opened 2016-05-20
Fix the project name in the readme header

xStrom:
From what I can gather, this is a typo that I fixed.


ShawnMilo:
Wow. Nice catch. Thanks!

tealeg:
Thanks @xStrom  - if you only knew the number of times I've mistyped that!

Issue #230, opened 2016-06-24
cell.HMerge


coolit:
There are some cells not merged, but get a value of 1 with Cell.HMerge.
I copied these cells and save as a new xlsx file, where the problem didn't occur again.
Sorry for not being able to upload the sample file, hoping this information can be of some help.

Issue #212, opened 2016-05-03
merge cells other cells also merged


zhoujianhe:
when we run sheetM.Cell(70, 0).VMerge = 1,cell(70,0) and cell(71,0) will merged,but Cell(7, 0) and cell(8,0) also merged.
when we run sheetM.Cell(59, 0).VMerge = 1,cell(59,0) and cell(60,0) will merged,but Cell(5, 0) and cell(6,0) also merged like this

PR #233, opened 2016-07-13
Fix prefix check for getExtent for merged cells

santsai:
When doing prefix checking in getExtent for merged cells, need to stick `:` to the reference,
otherwise getExtent can return wrong info for a non merged cell A1 if there is a merged cell at A1n.
this would possibly close these two issues: https://github.com/tealeg/xlsx/issues/230 https://github.com/tealeg/xlsx/issues/212


tealeg:
@santsai awesome, and thanks for the heads-up ;-)

PR #234, opened 2016-07-18
Panic: runtime error: slice bounds out of range

thieupham-wf:
### Problem / Feature  

This is the fix for the panic when a shared formula contains a defined name.  For example, SUM(D44:H44)*IM_A_DEFINED_NAME

### Reviewers  

Please review @tealeg Thanks!


tealeg:
@thieupham-wf cheers.

thieupham-wf:
Thanks! @tealeg

Issue #237, opened 2016-07-30
Row.WriteSlice panic


horsley:
reflect: call of reflect.Value.Elem on slice Value
goroutine 6 [running]:
net/http.(*conn).serve.func1(0xc820062300)
	/usr/local/go/src/net/http/server.go:1389 +0xc1
panic(0x4843c0, 0xc8201fe880)
	/usr/local/go/src/runtime/panic.go:426 +0x4e9
reflect.Value.Elem(0x3d14e0, 0xc8201fe860, 0x97, 0x0, 0x0, 0x0)
	/usr/local/go/src/reflect/value.go:735 +0x228
github.com/tealeg/xlsx.(*Row).WriteSlice(0xc82018aa80, 0x3d14e0, 0xc8201fe860, 0x2, 0xc8201fe860)
	/Users/horsley/gopath/src/github.com/tealeg/xlsx/write.go:18 +0x95

PR #238, opened 2016-07-30
fix slice check panic

horsley:


tealeg:
Hi @horsley, not the above failing tests.  It seems odd that you should to break the test cases to fix the issue you're seeing.

tealeg:
Note to self: fixes #237

tealeg:
Fixed by #239

PR #239, opened 2016-07-31
fix pointer to slice check panic on WriteSlice

santsai:
@tealeg I think this would be a right fix to #237 #238 
the panic is caused by calling Elem on non Ptr kind value,
this fix makes sure the value is a Ptr before calling Elem


tealeg:
@santsai perfect, thanks for you work!

PR #243, opened 2016-08-12
make WriteSlice and WriteStruct write some value using SetValue

klniu:
The commit make WriteSlice and WriteStruct write value with builtin format to general format, and make them support time.Time.

Add more types to SetValue and add test for cell.SetValue

Besides, I think Cell.SetInt, cell.SetInt64, cell.SetFloat should also use SetValue for setting values, so that the cell has a general format which can be recognized properly by spreadsheet program. I didn't do it in the commit because I'm not unsure if the change will affect other parts. It's my pleasure for doing that if you are sure it.

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -174,19 +174,7 @@ func (c *Cell) SetValue(n interface{}) {
 	case time.Time:
 		c.SetDateTime(n.(time.Time))
 		return
-	case int:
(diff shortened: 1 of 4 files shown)


tealeg:
@klniu excellent work!  Thanks very much.

I would be happy to see the other changes you describe.

Issue #179, opened 2015-11-03
Support freezing/locking rows/columns


epelc:
It would be really helpful if this supported freezing rows and columns(rows are more important).

See this for an example of freezing/locking rows/columns in excel.
https://support.office.com/en-us/article/Freeze-or-lock-rows-and-columns-08005001-fd23-4e2a-b370-77ab54d78e3f

stanim:
It should be done with SheetViews and Panes but it doesn't seem fully implemented yet, as the following doesn't seem to work:

```
 sheet.SheetViews = []xlsx.SheetView{
			xlsx.SheetView{
				Pane: &xlsx.Pane{
					XSplit:      1,
					YSplit:      1,
					ActivePane:  "bottomRight",
					TopLeftCell: "B2",
					State:       "frozen",
				},
			},
		}
```
See: https://godoc.org/github.com/tealeg/xlsx#Pane

epelc:
Thanks! I ended up creating my own xlsx writer package which implements this. iirc it wasn't implemented here.

stanim:
Is it available somewhere?
Op 10 jul. 2016 02:05 schreef "Ed Pelc" <notifications@github.com>:

epelc:
It isn't public now. But I might open source it.

On Sunday, July 10, 2016, Stani <notifications@github.com> wrote:

> Is it available somewhere?
> Op 10 jul. 2016 02:05 schreef "Ed Pelc" <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>:
>
> > Thanks! I ended up creating my own xlsx writer package which implements
> > this. iirc it wasn't implemented here.
> >
> > —
> > You are receiving this because you commented.
> > Reply to this email directly, view it on GitHub
> > <https://github.com/tealeg/xlsx/issues/179#issuecomment-231562701>, or
> mute
> > the thread
> > <
> https://github.com/notifications/unsubscribe/ACE9aZPccOC5HRKR-zouozVrwP3tKrd_ks5qUDc4gaJpZM4GbMmt
> >
> > .
> >
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/tealeg/xlsx/issues/179#issuecomment-231572241>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AE9qolPBJjTpaxXvefSzjrwo47xJtrD_ks5qUIetgaJpZM4GbMmt>
> .
[...]

PR #245, opened 2016-08-24
Fix SheetView Pane

francisbouvier:
Hi,

I've fix issue  #179 to support frozen pane.

Thanks for your lib, very useful.

Signed-off-by: Francis Bouvier <francis.bouvier@gmail.com>


tealeg:
@francisbouvier - I've merged this because it's a nice simple fix, but please in general add unit tests for functionality you add - it's the only way to ensure we don't regress in the future.

Issue #248, opened 2016-10-03
Sheet with single merged cell produces style="" in styles.xml


nieware:
I was having issues with opening an xlsx file in Excel Viewer and have narrowed it down to following code:

```
file := xlsx.NewFile()
sheet, _ := file.AddSheet("S1")
row := sheet.AddRow()
cell1 := row.AddCell()        
row.AddCell()        
file.Save("unmerged.xlsx")
cell1.Merge(1, 0)
file.Save("merged.xlsx")
```

"unmerged.xlsx" can be opened in Excel Viewer, "merged.xlsx" can't. Apparently this is due to the following XML fragment in xl/styles.xml:

```
<border><left style="none"></left><right style=""></right><top style="none"></top><bottom style="none"></bottom></border>
```

If I replace the style="" with style="none", it works fine. According to #211, this should already be fixed, but apparently there are cases where it still happens...

nieware:
I think I found the fix myself. In func (s *Sheet) handleMerged(), in lines 145-148:

    mainstyle.Border.Top = ""
    mainstyle.Border.Left = ""
    mainstyle.Border.Right = ""
    mainstyle.Border.Bottom = ""

I have replaced "" with "none" and voila, it works, and all tests still pass. Should I submit a pull request?

OneOfOne:
@nieware please do

mauleyzaola:
@nieware Shouldn't this issue be closed?

PR #250, opened 2016-10-14
Fix for issue #248

nieware:
This should fix issue #248 (merging cells produces empty style attrib). Also includes some cell bounds checks by user K1000000.


tealeg:
Thanks @nieware I think we're going to have some complaints about the "none" values, because it's come up before, but the spec supports it, and I'm happy to have this fix.

PR #254, opened 2016-10-26
Don't generate map on hot spot in code during iterations, precalculate

ReneKroon:
while profiling my application, a reasonable amount of time was spent creating an excel sheet. I did not find any big mistakes or optimization points, but noticed that this trivial inverse map is created often enough to show up in the profiling diagram. I therefore moved it to a static one.


tealeg:
@ReneKroon thanks - nice to see some work on performance.

Issue #259, opened 2016-11-14
XLSX file corrupts when saving the file without adding Worksheet


sathishbaburathinavel:
Following Lines of code corrupts the excel document. MS Excel expects at least one worksheet inside workbook, but the below code snippets save the file without adding worksheet to workbook.
```
func main() {
    var file *xlsx.File
    var err error
    file = xlsx.NewFile()
    err = file.Save("MyXLSXFile.xlsx")
    if err != nil {
        fmt.Printf(err.Error())
    }
}
```
**Issue Reproducing Scenario:**
* Execute the above code
* Open the MyXSLXFile.exe

**Obtained Result:** Document corrupted in MS Excel 2016
**Expected Result:** Document should not be saved.

sathishbaburathinavel:
Fixed this issue and requested the pull.
issue-259-XLSX file corrupts when saving the file without adding Worksheet in  the workbook"
Pull Request: [pull request 260](https://github.com/tealeg/xlsx/pull/260/commits)

PR #260, opened 2016-11-14
issue-259-XLSX file corrupts when saving the file without adding Work…

sathishbaburathinavel:
Fixed the issue by throwing the error if there is sheet in the workbook.
Now library throws error "Workbook must contains at-least one worksheet" to the user.

*Issue Link:*[Issue-259]( https://github.com/tealeg/xlsx/issues/259)

### DIFF
--- a/file.go
+++ b/file.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"errors"
 )
 
 // File is a high level structure providing a slice of Sheet structs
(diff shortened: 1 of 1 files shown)


file.go
```diff
@@ -205,6 +206,10 @@ func (f *File) MarshallParts() (map[string]string, error) {
 		f.styles = newXlsxStyleSheet(f.theme)
 	}
 	f.styles.reset()
+	if len(f.Sheets)==0 {
+		err:= errors.New("Workbook must contains atleast one worksheet")
```
tealeg: There's a space missing between "at" and "least".

ensonik:
Btw: Change wasn't backwards compatible. Broke a few of our builds (that create bogus sheets for testing).

I don't have any intelligent feedback or solutions; just thought I'd let you know in case you get some complaints.

PR #263, opened 2016-12-07
cell.String has two value.

p1gd0g:


tealeg:
Great, thanks @p1gd0g

Issue #253, opened 2016-10-22
File.ToSlice() fails due to merged cell has Number type and "shadowed" cell is empty 


regorov:
Hi tealeg! 
I've marked cell, on what fails.
[example.xlsx](https://github.com/tealeg/xlsx/files/545780/example.xlsx)

I guess, adding if{} block in the method below can resolve issue.

```
func (c *Cell) FormattedValue() (string, error) {
    if c.Value == "" {
        return "", nil
    }
```

What do you think?

Have a nice weekend!
Regards,
Robert

PR #265, opened 2016-12-07
Recover from error in file.ToSlice().

dizk:
Recover from strconv.NumError when cell.Value is "". Contains example .xlsx file, test and fix.


dizk:
This will probably also fix #253 😄 

Regards,
Didrik

tealeg:
@dizk Excellent fix.  Thanks for the good work!

PR #272, opened 2017-01-16
Func. appending an existing sheet to a file

tokyis:
In cases where Sheet(s) is created before File

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -114,11 +114,11 @@ func (c *Cell) SetFloatWithFormat(n float64, format string) {
 
 var timeLocationUTC, _ = time.LoadLocation("UTC")
 
-func timeToUTCTime(t time.Time) time.Time {
+func TimeToUTCTime(t time.Time) time.Time {
 	return time.Date(t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second(), t.Nanosecond(), timeLocationUTC)
 }
 
-func timeToExcelTime(t time.Time) float64 {
+func TimeToExcelTime(t time.Time) float64 {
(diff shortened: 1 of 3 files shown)


cell.go
```diff
@@ -114,21 +114,21 @@ func (c *Cell) SetFloatWithFormat(n float64, format string) {
 
 var timeLocationUTC, _ = time.LoadLocation("UTC")
 
-func timeToUTCTime(t time.Time) time.Time {
+func TimeToUTCTime(t time.Time) time.Time {
```
tealeg: I'm not sure why you needed to change this, the only consumer is in the same module below.

tokyis: My bad, changes from d5dc146f4ea75e6b34d1a79324cedb22dd708157 on should have been in another branch/pull request altogether, I didn't expect this pull request to show them. 
It's the only way I found to set custom date formats using SetDateTimeWithFormat as a user that would be consistent with what's already there. Granted those 2 if exported could use a move to a more "timely" package. ^^
I also saw another pull request for that which looked more thought through, those changes were meant to stay on my repo for this one time I needed them.
Feel free to ignore those particular changes.

tokyis:
My bad, changes from d5dc146f4ea75e6b34d1a79324cedb22dd708157 on should have been in another branch/pull request altogether, I didn't expect this pull request to show them. 
It's the only way I found to set custom date formats using SetDateTimeWithFormat as a user that would be consistent with what's already there. Granted those 2 if exported could use a move to a more "timely" package. ^^
I also saw another pull request for that which looked more thought through, those changes were meant to stay on my repo for this one time I needed them.
Feel free to ignore those particular changes.

Issue #273, opened 2017-01-27
Is there a way to output times and dates in a certain format?


ds0nt:
I have a time.Time and I want it to export into excel showing in a different timezone.

A quick glance at the code gives me the impression that it converts everything into UTC and that's it.

Is there something I'm missing?

I'd be happy to make a PR with changes if neccessary

```go

func timeToUTCTime(t time.Time) time.Time {
	return time.Date(t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second(), t.Nanosecond(), timeLocationUTC)
}

func timeToExcelTime(t time.Time) float64 {
	return float64(t.Unix())/86400.0 + 25569.0
}

// SetDate sets the value of a cell to a float.
func (c *Cell) SetDate(t time.Time) {
	c.SetDateTimeWithFormat(float64(int64(timeToExcelTime(timeToUTCTime(t)))), builtInNumFmt[14])
}

func (c *Cell) SetDateTime(t time.Time) {
	c.SetDateTimeWithFormat(timeToExcelTime(timeToUTCTime(t)), builtInNumFmt[22])
}

func (c *Cell) SetDateTimeWithFormat(n float64, format string) {
	c.Value = strconv.FormatFloat(n, 'f', -1, 64)
	c.NumFmt = format
	c.formula = ""
	c.cellType = CellTypeDate
}
```

tealeg:
@ds0nt your PR addresses this (thanks!) - it just needs some unit tests and I'll be happy to merge it.

ds0nt:
Alright, I'll see if I get to it this weekend.

Thanks!

PR #274, opened 2017-01-27
Can output times and dates with custom formats and timezones.

ds0nt:
https://github.com/tealeg/xlsx/issues/273


tealeg:
Sorry for the late review.  There should be unit tests to cover this new functionality, otherwise we can't be sure that we don't regress following future changes.  Any chance you could add some?

ds0nt:
I'll try and give it a look this weekend, depending on life-stuff.

Thanks!

ds0nt:
I've updated the tests. Can you take a look and see if you think they reasonably test all cases? I think there could be some corner-cases I am missing, but here's an attempt.

tealeg:
@ds0nt sorry for the delay here.  I've been distracted looking for a job!  Anyhow, this looks good now, I'm just resolving a conflict for you and rerunning the tests.  Should land shortly.

Issue #278, opened 2017-02-17
Reading xlsx file with more but one sheet throws panic


the42:
As of Golang 1.8, Reading a XLSX file with more but one Sheet fails. The issue seems to have been foreseen somehow,

https://github.com/tealeg/xlsx/blob/master/lib.go#L659

but throws now a Panic at this channel write

ShawnMilo:
I'm not reproducing this with a file with three worksheets, on Go 1.8 final. @the42, please send an XLSX file I can use to work on this. Thanks!

HeinOldewage:
A data race exists surrounding that channel. 
I have been away of it for some time, but the only time it should panic is if a workbook has two sheets in, with both containing errors. 
Some timing changes might have happened in go1.8 making the case more likely.  
Reading the workbooks should be rewritten to a more idiomatic way. This will be some work though.

Example test case that breaks:
````
go test -race -tags=fuzzy -test.seed=1487522509
````

I will see if I can find a nice refactor for the code.

the42:
Coming back to your request to duplicate the error / provide an XLSX-file. I think what raised my issue is somehow related to the already fixed #275:

- I created an XLSX-file using Libreoffice - no issues reading the file with tealeg/xlsx
- I opened the file using Excel 2010 (on Windows), added another sheet, added data, saved the file

I guess what has happened then was that this file triggered for both sheets the error in #275 which (as you identified) triggers the race condition.

The switch to Go1.8 was likely only coincidence.

tealeg:
@the42 thanks for reporting the bug.  I'll close this for now as it seems related to the two issues we've recently merged fixes for.  Don't hesitate to reopen the issue should it reoccur.

tealeg:
@HeinOldewage pretty sure the combination of your fix and the one I did on Friday have cleared this up.  Thanks a lot for your work!

tealeg:
@ShawnMilo looks like we've killed this one.  Thanks for monitoring this stuff!

PR #280, opened 2017-02-19
Fixed the datarace surrounding the channel. 

HeinOldewage:
This was done by moving the channel close to the writer goroutine (as is idiomatic). readSheetFromFile
now returns an error to stop the writer routine from trying to read more
sheets if one contains an error (this will prevent goroutines from
leaking due to readers that stop reading on receiving an error).

This might fix issue #278.


tealeg:
@HeinOldewage wonderful, thanks very much for this!

PR #281, opened 2017-02-21
Inline string cells

jfyne:
Hi thanks for the library it is great!

The project I work on receives a lot of random excel files, I came across a file the other day where upon reading it contained absolutely no strings. I dug into the file itself and saw that it appeared the strings were living in ```<is>``` elements, probably malformed.

I added this code to fix my specific problem, to me it feels like a bit of a hack. Is there a better way to handle this case?


tealeg:
@jfyne this isn't malformed at all, it exists in the ECMA documentation for the Office Open XML Spreadsheet ML format.  I don't know exactly why it's being used in your case, I've not seen it before, but it's completely valid.  If you feel like tidying up your pull request (remove the references to "malformed")  then I'd be happy to review it for merging.

jfyne:
@tealeg Ok all references should be removed to malformed. Thanks

jfyne:
@tealeg bump

tealeg:
@jfyne sorry for the epic delay - I lost track of this.  Still, excellent work, thanks!

PR #287, opened 2017-03-13
Properly escape `xlsxNumFmt.FormatCode` in `Marshal()`

g7r:


g7r:
The latter two commits arrived here unintentionally. I haven't found a way to remove them from pull request, so I'll create new one.

PR #296, opened 2017-04-14
Add function "SetHeight".

zzc-tongji:
When setting the height of a row equal to another, I have no idea but to use:
`row.SetHeightCM(sheet.Rows[0].Height / 28.3464567)`

So I think a function of setting row's height directly is necessary.


tealeg:
@zzc-tongji thanks.

PR #309, opened 2017-06-05
Add Sheet length check

doatesy:
Returns error if new sheet name is >= 31 charaters long


tealeg:
@doatesy good catch!  Please could you add a unit test that checks this, that way we can avoid regressions in the future.

Issue #305, opened 2017-05-22
Proposal: add method for reading structs from rows


thewraven:
### Problem

Reading structs from a row which has a well-known structure is a bit repetitive, because you will have to iter over the cells, check for the desired type to be obtained, checking for errors and filling the struct.

### Current approach

If you want to read a whole file (or a single sheet) and convert it to a slice of structs, you have to take any of these two paths:

- Use [FileToSlice](https://godoc.org/github.com/tealeg/xlsx#FileToSlice) iter over the slices and handle the type-casting for yourself
- Iter over the Sheets of the file, then over the rows and filling the structs by yourself

### Proposal

Currently we have a method in *Row to [Write a struct](https://godoc.org/github.com/tealeg/xlsx#Row.WriteStruct) , so maybe we can define a symmetric method like `(*Row).ReadStruct(interface{}) error` that receives a struct pointer to be filled. 

To know the relation column <-> field we can use struct tags that defines the index of the cell to be dumped in the field, like so:

```go
type structTest struct {
	IntVal     int     `xlsx:"0"`
	StringVal  string  `xlsx:"1"`
	FloatVal   float64 `xlsx:"2"`
	IgnoredVal int     `xlsx:"-"`
	BoolVal    bool    `xlsx:"4"`
}
```

I'm wondering if would be better to use a name for the column, but I sincerely don't know how this would work.

I have a very simple implementation of the proposal using the tags in a forked repo:

[Implementation commit](https://github.com/thewraven/xlsx/commit/3c4837368e1935c365593ae769080765df109acb)

Comments are welcome!

tealeg:
Hi @thewraven  - this is a fine idea.  My first thought is that if we use annotations on the struct in that manner then these functions should be Marshal() and Unmarshal() as per the ones in the go standard library encoding/ packages.    If you did a Marshal() function, I think it would be nice to allow setting the numFmt from the struct annotation.

thewraven:
Maybe we can use both interfaces for marshalling/unmarshalling + xlsx tags?

So, if I want to make a custom marshal, I should implement the following interfaces
(it's a proposal of the signature):

```go
type Marshaler interface {
    MarshalXLSX(*Row) error
}

type Unmarshaler interface {
    UnmarshalXLSX(*Row) error
}
```
Am I missing something?

tealeg:
The marshal signature would typically be more like

`type XLSXMarshaler interface {
        MarshalXLSX(v interface{}) (* Row, err error)
[...]

PR #311, opened 2017-06-10
Read struct from row based on xlsx tags.

thewraven:
Supported types: bool,int{8,16,32,64},float64,string, time.Time

Also custom deserialization is supported via XSLXUnmarshaler interface


thewraven:
This is an implementation for issue #305

tealeg:
Hi @thewraven this is looking interesting.  Can you address the failing test above?

thewraven:
Sure, I'll submit the change when its ready :+1:

tealeg:
@thewraven great work.  I'll merge this is as is.  I might eventually create an "examples" directory and move your example code in there.

Issue #313, opened 2017-07-10
Microsoft OOXML?


0x4139:
Hello, i have an issue with an excel file, i'm not sure what exactly happens but the behavior is as follows:
`file test.xlsx  -> test.xlsx: Microsoft OOXML`
when i try to parse it with the xlsx package i can see only the top row.
when i save it using excel
`file test.xlsx - > Microsoft Excel 2007+`
if i try to reparse the data i see it as i should the rows are there.

Please advise.

0x4139:
ok, so the issue is related to the OOXML format that sometimes comes with only one dimension even if it has rows, the fix was to fallback to counting from the worksheet when we don't have 2 dimensions.

PR #314, opened 2017-07-10
Microsoft OOXML sometimes comes without the floor dimension ref

0x4139:
Microsoft OOXML sometimes comes without the floor dimention ref, so we need to fallback to the calculateMaxMinFromWorkSheet
Fixes:
#313


tealeg:
@0x4139 good catch!

PR #321, opened 2017-07-22
WIP: MergeCells: Optimization for large sheets.

### LINKED ISSUE #312: 100% CPU & 530G VIRT memory used when opening excel file (specific file linked with demo)
## My Understanding of the Problem

I'm experiencing issues with opening a large excel file I have, which has been generated by https://microstrategy.gmu.edu (I'm fairly certain Microstrategy uses sheet.js). While the file size itself is only 2.6 MB, it does have around a half a million cells populated with data (this sheet contains every section of every course being taught at my university along with information about each).

The exact spreadsheet I am trying to open can be found at https://github.com/ocelotsloth/schedules/blob/master/dist/exampleFall2017.xlsx. 

When I try to open this file, the program hangs, consumes 100% CPU, and allocates itself a truly absurd amount of memory (or swap? To be honest I'm still not 100% clear on what VIRT is):

hr9wn6d

## Research I have done so far

I have done some research already and do not think it is related to either #109 or #114. This file is significantly smaller in filesize, and as explained below, the offending commit doesn't appear to be directly related to opening files (though I could very well be wrong on that second assertion).

This is a cross-platform issue that I have tested by compiling against Windows, Darwin, and Linux.

I performed a `git bisect` and was able to narrow down the offending commit to e07ef57987388ad2510dac5fa2b5dde5945b1b31, which was authored by @tealeg in October of 2015 (sorry to bring up such a distant memory). I've added the commit message below for quick reference:

    commit e07ef57987388ad2510dac5fa2b5dde5945b1b31
    Author: Geoffrey J. Teale <geoffrey.teale@canonical.com>
    Date:   Fri Oct 2 23:19:48 2015 +0200
    
        Populate the VMerge and HMerge properties of the cell based on the xlsxWorksheet.MergeCells.

This commit adds in some actions dealing with Merge Cells, but since I was more concerned with graduating high school than learning Golang at the time, I couldn't tell you the full extent of what it does and also don't know how fits in with other work being done at the time.

## A helpful repository to explore and test

Since the project I'm using this library in is actually my way of picking up and learning Golang, I'm well aware that I'm not really going to be of much further use to try and take a stab at the exact problem here. In lieu of that kind of help, I've gone ahead and made a simple program to clearly demonstrate the issue. 

`go get github.com/ocelotsloth/xlsx_issue_demo` should pull down both the demo program as well as the excel spreadsheet I am working with.

ocelotsloth:
This patch fixes #312 (https://github.com/tealeg/xlsx/issues/312)
by removing an unneeded loop in `xmlWorksheet.getExtent` which,
for large excel sheets, iterates an excessive amount searching for
the desired merge cell via string operations and a for loop.

By adding a map to the `xlsxMergeCells` struct which stores the same
data as the `Cells` the patch preserves backwards compatibility for
any users of the library making use of that struct while also
allowing the `getExtent` function to simply use the cell reference
as a key to locate the appropriate cell.

This does not change the API in any way besides the addition of the
field to the `xlsxMergeCells` struct, so the existing test cases
should cover this.

Signed-off-by: Mark Stenglein <mark@stengle.in>

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -334,6 +334,7 @@ func (s *Sheet) makeXLSXSheet(refTable *RefTable, styles *xlsxStyleSheet) *xlsxW
 					worksheet.MergeCells = &xlsxMergeCells{}
 				}
 				worksheet.MergeCells.Cells = append(worksheet.MergeCells.Cells, mc)
+				worksheet.MergeCells.CellsMap[start] = mc
 			}
 		}
 		xSheet.Row = append(xSheet.Row, xRow)
--- a/sheet.go
+++ b/sheet.go
@@ -330,9 +330,6 @@ func (s *Sheet) makeXLSXSheet(refTable *RefTable, styles *xlsxStyleSheet) *xlsxW
 				endrow := r + cell.VMerge + 1
 				end := fmt.Sprintf("%s%d", numericToLetters(endcol), endrow)
 				mc.Ref = start + ":" + end
-				if worksheet.MergeCells == nil {
-					worksheet.MergeCells = &xlsxMergeCells{}
-				}
 				worksheet.MergeCells.Cells = append(worksheet.MergeCells.Cells, mc)
 				worksheet.MergeCells.CellsMap[start] = mc
 			}
--- a/xmlWorkbook.go
+++ b/xmlWorkbook.go
@@ -5,6 +5,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"strings"
 )
 
 const (
@@ -198,6 +199,11 @@ func getWorksheetFromSheet(sheet xlsxSheet, worksheets map[string]*zip.File, she
 	error = decoder.Decode(worksheet)
 	if error != nil {
 		return nil, error
+	}
+	worksheet.MergeCells.CellsMap = make(map[string]xlsxMergeCell)
+	for _, cell := range worksheet.MergeCells.Cells {
+		cellRefs := strings.Split(cell.Ref, ":")
+		worksheet.MergeCells.CellsMap[cellRefs[0]] = cell
 	}
 	return worksheet, nil
 }
--- a/xmlWorksheet.go
+++ b/xmlWorksheet.go
@@ -246,9 +246,10 @@ type xlsxMergeCell struct {
 }
 
 type xlsxMergeCells struct {
-	XMLName xml.Name        //`xml:"mergeCells,omitempty"`
-	Count   int             `xml:"count,attr,omitempty"`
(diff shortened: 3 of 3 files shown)


ocelotsloth:
If you compare commit's across github.com/ocelotsloth/xlsx_issue_demo you will notice a significant improvement.

Rather than taking 50 minutes, this patch speeds up the process to about 5 seconds on my computer (with the large file) and still noticeably faster with smaller excel sheets with merge cells.

I'm still curious to try and optimize this more, but this is a simple fix which removes the CPU bottleneck.

ocelotsloth:
Hmm...guess I need to take another look. :worried:

ocelotsloth:
Ok, so I think what's happening is that the import is not writing the values when the xml parser does it's job. I'm not really surprised by this, but I'm having a hard time figuring out where to add something to populate the new field. Actually remembering to run `go test` would have also been a good start.

Is there a good place to process the data from a loaded file once the xml processor is done?

ocelotsloth:
If you take a look at github.com/ocelotsloth/xlsx_issue_demo you will be able to see that this does actually work, and **much** faster at that.

I'm not positive on how best to approach the issues with the nil pointers though, and would be appreciative of a point in the right direction.

tealeg:
Hey @ocelotsloth sorry for being silent for so long.  I'll try and find a moment to look at this, I'm not ignoring you I promise!

ryho:
Hm, I can't get the demo XLSX file to open in Numbers at all either.

You need to add `if worksheet.MergeCells != nil {` before line `xmlWorkbook.go:203`. If you do that, there will be only three tests failing, and it's because they test merge cells without initializing `CellsMap`.

However, adding the `CellsMap` directly to xlsxMergeCells is pretty hacky, because everything extra should live its own Foo struct, and the xlsxFoo struct should only contain the information from the XLSX spec, but I can see that makes this code need some refactoring. I think the only reason that your change isn't causing problems with the XML marshaling is because maps can't be marshaled to XML, otherwise your map would be showing up in the exported XLSX file data. 

FYI I was able to figure out the line where the exception was occurring by searching for `recover()` and removing it's usage.

ocelotsloth:
Hey @tealeg I'm just looking through stale PRs of mine and noticed this was still open. Has this issue been addressed in the last few years? If not I'd be happy to revisit this contribution.

tealeg:
@ocelotsloth Some years ago now I had to step back from this project as I had too much on my plate, and I wasn't using XLSX.  @ryho has been running things around here since then, so it's probably best to ask him.

tealeg:
@ocelotsloth now that I'm back working on xlsx a bit, I took the time to fix this PR and merge it.  Thanks for your work.

Issue #323, opened 2017-07-24
Streamed XLSX writing


ryho:
Would you be interested in adding features that allow streamed writing? I've written a wrapper around your library that allows XLSX files to be streamed to an io.Writer as they are being created: https://github.com/ryho/excel_stream
Because the project that I am currently working on currently streams to CSVs, the fastest way for me to proceed was to created a streamed XLSX library rather than refactor the existing code. This is sort of a niche feature and would make the library's API a little bit complicated since there would be two ways to use it, and this would add code maintenance costs to your plate, but I think it could be useful for others if it was integrated into the library rather than a wrapper.

tealeg:
Hi @ryho,

In principal, yes, streamed version is something I've thought about for a long time.  I'm quite interested in the approach you took, as I've always assumed it'd require a rewrite of a lot of the logic.

Initial thoughts: we could incorporate it as a sub package (so you'd import github.com/tealeg/xlsx/stream ).

PR #324, opened 2017-07-25
Stream Writing, Partial Reads, Improved FormattedValue() in General mode

ryho:
- https://github.com/tealeg/xlsx/issues/323 Supports stream writing with an API that matches the golang CSV reader so that these can be passed as the same interface. It also supports setting the formatting type of columns.
- https://github.com/tealeg/xlsx/issues/338 https://github.com/ryho/xlsx/pull/2 All public functions now support specifying the number of rows to be read so you can partial read XLSX files to quickly peek at the first few rows of large files.
- https://github.com/tealeg/xlsx/issues/334 https://github.com/ryho/xlsx/pull/1 Cell.FormattedValue() is more correct in the General case when the value is a number.
- https://github.com/tealeg/xlsx/issues/337 https://github.com/ryho/xlsx/pull/3 The function that turns column indexes into letters is now public. Previously you could only turn a cell into an Excel coordinate, but not just a column number. Also the function that turns coordinates into Excel strings is more widely used, the behavior was duplicated throughout the library.
- https://github.com/ryho/xlsx/pull/4 I went back and improved the tests for partial reads so they are less flaky.


ryho:
Have you considered removing the golang "tip" test? The other tests take less than 30 seconds, but that test run takes 2 1/2 minutes. And I really can't imagine you editing the library's code in response to go 1.5-1.8 passing but go tip failing. If that happened, you would just remove the tip test.

How important are go 1.5 and 1.6? It seems that only 1.7 and 1.8 are supported: https://golang.org/dl/

ryho:
`gimme stable` is currently broken when using travis: https://github.com/travis-ci/gimme/issues/93
`gimme master` takes 2.5 mins
`gimme tip` takes 2.5 mins
Also `gimme 1.7` and `gimme 1.8` are providing go1.7.0 and go1.8.0 not go1.7.X and go1.8.X, but they take 18, and 30 seconds.

ryho:
PTAL, sorry for so much noise.

I edited the test configuration a little bit:
* `1.5` and `1.6` fail because my tests use new Go testing features, but those versions of Go are unsupported anyway, so I think we should remove them: https://golang.org/dl/
* `tip` takes 2.5 mins, which is way longer than all the other versions, and I really can't imagine someone editing the library's code in response to stable versions passing but tip failing. If that happened, one would probably just remove the tip test.
* I tried `stable` but  is currently broken when using Travis and returns go1.6: travis-ci/gimme#93
* I tried `master` but it also takes 2.5 mins
* We can update `1.7` and `1.8` to be `1.7.x` and `1.8.x` to get the latest of those versions
* `go get -d -t -v ./... && go build -v ./...` is no longer necessary and causes a redundant build
* `go test` has to be called once for each package with the `-coverprofile` flag

stream/excel_stream_test.go
```diff
@@ -0,0 +1,418 @@
+package excel_stream
```
tealeg: This package name seems wrong.  XLSX is an open format implemented by multiple pieces of software, not just Excel.  Given that it's already going to be prefixed with github.com/tealeg/xlsx why don't we just call this "stream" ?

stream/excel_stream_test.go
```diff
@@ -0,0 +1,418 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/tealeg/xlsx"
+)
+
+func TestXlsxStreamWrite(t *testing.T) {
+	// When shouldMakeRealFiles is set to true this test will make actual XLSX files in the file system.
+	// This is useful to ensure files open in Excel, Numbers, Google Docs, etc.
+	// In case of issues you can use "Open XML SDK 2.5" to diagnose issues in generated XLSX files:
```
tealeg: This tip would be good to add to the top level README.org file.

stream/stream_file.go
```diff
@@ -0,0 +1,164 @@
+package excel_stream
```
tealeg: As per above, I'd prefer just "stream".

stream/stream_file_builder.go
```diff
@@ -0,0 +1,215 @@
+// Authors: Ryan Hollis (ryanh@)
+
+// The purpose of the StreamFile library is to allow streamed writing of XLSX files.
+// It relies heavily on the XLSX library (github.com/tealeg/xlsx).
```
tealeg: Don't think we need to say that anymore if it's in the repository ;-)

stream/stream_file_builder.go
```diff
@@ -0,0 +1,215 @@
...
+// Directions:
+// 1. Create a StreamFileBuilder with NewStreamFileBuilder() or NewStreamFileBuilderForPath().
+// 2. Add the sheets and their first row of data by calling AddSheet().
+// 3. Call Build() to get a StreamFile. Once built, all functions on the builder will return an error.
+// 4. Write to the StreamFile with WriteRow(). Writes begin on the first sheet. New rows are always written and flushed
+// to the io. All rows written to the same sheet must have the same number of cells as the header provided when the sheet
+// was created or an error will be returned.
+// 5. Call NextSheet() to proceed to the next sheet. Once NextSheet() is called, the previous sheet can not be edited.
+// 6. Call Close() to finish.
+
```
tealeg: These directions should be move to (or duplicated in) a new section in the top-level README.org file.  If you're not comfortable with the org-mode formatting let me know and I'll be happy to do that part myself.   Probably we should also indicate that this feature is new and not yet feature complete.

stream/stream_file_builder.go
```diff
@@ -0,0 +1,215 @@
...
+// Future work suggestions:
+// Currently the only supported cell type is string, since the main reason this library was written was to prevent
+// strings from being interpreted as numbers. It would be nice to have support for numbers and money so that the exported
+// files could better take advantage of Excel's features.
+// All text is written with the same text style. Support for additional text styles could be added to highlight certain
+// data in the file.
+// The current default style uses fonts that are not on Macs by default so opening the XLSX files in Numbers causes a
+// pop up that says there are missing fonts. The font could be changed to something that is usually found on Mac and PC.
+
```
tealeg: It would be nice to have these as issues on the Github repo.  Carefully tagged so you don't lose them ;-)

stream/stream_file_builder.go
```diff
@@ -0,0 +1,215 @@
...
+package excel_stream
```
tealeg: Package name again.

stream/excel_stream_test.go
ryho: Oh, I meant to change it. I actually never realized that the package name in Go doesn't have to match the folder name.

ryho:
I've created issues for my future work comment, but I cannot tag or assign them because I don't have the permission:
https://github.com/tealeg/xlsx/issues/325
https://github.com/tealeg/xlsx/issues/326
https://github.com/tealeg/xlsx/issues/327

ryho:
I'm wondering if we should actually put this code in the same package. My IDE is complaining because the names of things inside the stream package start with the word stream (stream.StreamFile, stream.StreamFileBuilder). I don't like the idea of there being two things in the library with the name File (xlsx.File and stream.File) and I also don't like the idea of the word stream being used twice (stream.StreamFile).
I also think that in the future we may run into problems as we try to more tightly integrate the streaming logic into the rest of the code. For example in https://github.com/tealeg/xlsx/issues/325, as it stands, I can't actually use the `xlsx.xlsxRow` and `xlsx.xlsxCell` structs in the `stream` package.

ryho:
During the course of using this library, I've realized it will be much easier if this has the same interface as Go's `encoding/csv.Writer`. I've update the names accordingly.

ryho:
I've discovered the the files made by the stream writer are HUGE. I can export the same data as 3.5 MB CSV file or a 64MB XLSX file. I'm going to do a some more work on this to get the file size down by using shared strings and omitting empty cells.

ryho:
I eventually discovered that the golang zip library can be stream zip files while also compressing them. The reason I thought that this was not the case, is that the zip writer waits until it has 20k bytes until it starts streaming, regardless of how often you call Flush(). The library now zips the files, and the file sizes are much more reasonable.

The stream library now has an additional feature, it can mark columns and cells as type text. This was a blocker for our release of using this streaming library.

I've also decided to keep the stream logic inside the xlsx package, this will allow it to share private code so that it can do things in a less hacky way. Once this gets merged, I'll work on those changes as well.

PR #331, opened 2017-09-01
avoiding file description leak

1046102779:


lib.go
```diff
@@ -712,6 +712,11 @@ func readSheetsFromZipFile(f *zip.File, file *File, sheetXMLMap map[string]strin
 	if err != nil {
 		return nil, nil, err
 	}
+	defer func() {
+		if err = rc.Close(); err != nil {
+			return nil, nil, err
```
tealeg: You can't return the outer functions results from a deferred function like this.  You instead could try and do the close at the return points below.

1046102779:
thks

PR #341, opened 2017-10-04
added basic support for reading cell comments

nad2000:
I needed to access the cell comments. I have added a simple comment reading with reference to the commented cell, author, and the text value.

### DIFF
--- a/comments.go
+++ b/comments.go
@@ -1,0 +1,24 @@
+package xlsx
+
+import (
+	"strconv"
+)
+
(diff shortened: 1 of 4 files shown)


nad2000:
need further improvement. it turns out that there is a separate commentN.xml file per sheet. Need to link comments to respective sheet.

nad2000:
@tealeg, any suggestions how to better and where (Sheet or File) the comments? tia, Rad

PR #344, opened 2017-11-10
fix bug "no read alignment property RotationText"

presto78:
When reading an Excel file, the property TextRotation is not used.
This request fix it.


xmlStyle.go
```diff
@@ -211,8 +211,9 @@ func (styles *xlsxStyleSheet) getStyle(styleIndex int) *Style {
 			style.Alignment.Vertical = xf.Alignment.Vertical
 		}
 		style.Alignment.WrapText = xf.Alignment.WrapText
-
-		styles.Lock()
+        style.Alignment.TextRotation=xf.Alignment.TextRotation
+		
+        styles.Lock()
```
ryho: Can you fix the indentation so that it matches the rest of the block that it is in?

presto78: Ok! It's completed.

PR #345, opened 2017-11-13
Update README.org

ykocaman:
Reading XLSX file demo gave an error like this.
"multiple-value cell.String() in single-value context"


tealeg:
@ykocaman thanks!

PR #349, opened 2017-11-27
Fix SetColWidth,support setting width of any col

DennisMao:
Before:
The program will panic when set the width of cols which has not been initialized by AddCell.
Problem:
SetColWidth does not support insert cols automatically.
Now:
It supports set the width of any column,both single and multiple columns.


tealeg:
This PR actually missed the point that handling of the `<col>` element is currently quite incorrect.  It needs reworking completely.  I'm closing this because it's old, rotted and #474 represents a better fix.

PR #358, opened 2018-02-01
Stream writing with custom styles and types per row

viniciusps2:
Hello everyone!

I'm from NodeJS world, and we can't generate large XLSX files (500K+) in Javascript with high performance.

I was looking for a library that could do using stream, then I found tealeg/xlsx! 

But I needed to insert cells with style as stream, and at the moment it is only possible to set the type to collumns, before call the build method. 

I needed to create custom headers and rows with styles and types, and choose different styles and types per row. 

So I changed a few lines of code to achieve this. Here, a small example with the commented changes:

```
	// error handling was omitted for brevity
	file, err := xlsx.NewStreamFileBuilderForPath("stream.xlsx")

	// I allow AddSheet without headers (2nd parameter) and cellTypes (3rd parameter)
	err = file.AddSheet("sheetName1", nil, nil)

	boldStyle := xlsx.NewStyle()
	boldFont := xlsx.NewFont(10, "Arial")
	boldFont.Bold = true
	boldStyle.Font = *boldFont
	boldStyle.ApplyFont = true

	// I expose Sheet to allow insert and edit Cells, I used it to define custom headers.
	file.Sheet.AddRow().WriteSlice(&[]string{"a", "b", "c"}, -1)
	file.Sheet.Cell(0, 0).SetStyle(boldStyle)

	// Creating a style to be used in .WriteWithStyleFmtIds (when processing stream)
	redStyle := xlsx.NewStyle()
	font := xlsx.NewFont(10, "Arial")
	font.Color = "FFFF0000"
	redStyle.Font = *font
	redStyle.ApplyFont = true
	redStyleId := file.AddStyle(redStyle, xlsx.CellTypeNumeric)

	// Define style and cellType for colls and store in redStyleFmtIds
	redStyleFmtIds := &[]xlsx.StyleFmtId{
		{ redStyleId, xlsx.CellTypeNumeric },
		{ redStyleId, xlsx.CellTypeNumeric },
		{ redStyleId, xlsx.CellTypeString },
	}

	streamFile, err := file.Build()

	// Write row without style and types
	err = streamFile.Write([]string{"1", "2", "a string"})
	
	// Write row with style and use redStyleFmtIds
	err = streamFile.WriteWithStyleFmtIds([]string{"3", "4", "other string"}, redStyleFmtIds)

	err = streamFile.Close()
```

Please let me know if it is convenient to have some example of Stream in README, so I can write it.

### DIFF
--- a/file.go
+++ b/file.go
@@ -226,6 +226,14 @@ func replaceRelationshipsNameSpace(workbookMarshal string) string {
 	return strings.Replace(newWorkbook, oldXmlns, newXmlns, 1)
 }
 
+func (f *File) StyleSheet() *xlsxStyleSheet  {
+	if f.styles == nil {
+		f.styles = newXlsxStyleSheet(f.theme)
+		f.styles.reset()
+	}
+	return f.styles
+}
+
 // Construct a map of file name to XML content representing the file
 // in terms of the structure of an XLSX file.
 func (f *File) MarshallParts() (map[string]string, error) {
@@ -249,10 +257,8 @@ func (f *File) MarshallParts() (map[string]string, error) {
 	workbook = f.makeWorkbook()
 	sheetIndex := 1
 
-	if f.styles == nil {
-		f.styles = newXlsxStyleSheet(f.theme)
-	}
-	f.styles.reset()
+	f.StyleSheet()
+
 	if len(f.Sheets) == 0 {
 		err := errors.New("Workbook must contains atleast one worksheet")
 		return nil, err
--- a/stream_file.go
+++ b/stream_file.go
@@ -18,6 +18,11 @@ type StreamFile struct {
 	err            error
 }
 
+type StyleFmtId struct {
+	StyleId int
+	CellType CellType
+}
+
 type streamSheet struct {
 	// sheetIndex is the XLSX sheet index, which starts at 1
 	index int
@@ -43,7 +48,19 @@ func (sf *StreamFile) Write(cells []string) error {
 	if sf.err != nil {
 		return sf.err
 	}
-	err := sf.write(cells)
+	err := sf.write(cells, nil)
+	if err != nil {
+		sf.err = err
+		return err
+	}
+	return sf.zipWriter.Flush()
+}
+
+func (sf *StreamFile) WriteWithStyleFmtIds(cells []string, styleFmtIds *[]StyleFmtId) error {
+	if sf.err != nil {
+		return sf.err
+	}
+	err := sf.write(cells, styleFmtIds)
 	if err != nil {
 		sf.err = err
 		return err
@@ -56,7 +73,7 @@ func (sf *StreamFile) WriteAll(records [][]string) error {
 		return sf.err
 	}
 	for _, row := range records {
-		err := sf.write(row)
+		err := sf.write(row, nil)
 		if err != nil {
 			sf.err = err
 			return err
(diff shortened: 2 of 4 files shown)


ryho:
I'm glad someone is going else to use the streaming features I added!
I like everything, except I'm hesitant about exposing the Sheet field. I think that will let users do things they shouldn't do when using streaming, and it doesn't make it super clear what you're allowed to do.
I do like everything else though. Could you make an `AddSheetWithStyle` that's similar to the `AddSheet` function except it takes IDs returned from `AddStyle`? I think if you made that, the Sheet wouldn't need to be exposed.

viniciusps2:
I forget to congratulate you on this feature, it is very fast! My entire stack is Node.JS, so I call a GoLang binary and send data to stdint, it processes with this library, and it gets the XLSX from stdout, so Node.JS can compress it, and send to user in the same processing stream (one HTTP request). All this runs in less than 5 minutes.

I exposed Sheet because I need to customize the headers and set column width, for example, with this implementation I can call: "file.Sheet.SetColWidth(0, 1, 2)" before Build().

But I can implement a method to expose only this feature, for example, a "file.SetColWidth(0, 1, 2)" that must be called only before .Build(). What do you think?

ryho:
BTW, there is no need to compress the files in your Node.JS layer. The Excel files made by this library  are already compressed using gzip, which is allowed per the XLSX spec.

stream_file.go
```diff
@@ -65,36 +82,49 @@ func (sf *StreamFile) WriteAll(records [][]string) error {
 	return sf.zipWriter.Flush()
 }
 
-func (sf *StreamFile) write(cells []string) error {
+// documentation for the c.t (cell.Type) attribute:
```
ryho: It is probably worth mentioning in this comment that specifying "Shared String", won't make a shared string since it is impossible during streaming.

stream_file.go
```diff
@@ -65,36 +82,49 @@ func (sf *StreamFile) WriteAll(records [][]string) error {
...
-		cellOpen += `><is><t>`
-		cellClose := `</t></is></c>`
+		
+		cellCoordinate := GetCellIDStringFromCoords(colIndex, sf.currentSheet.rowCount-1)
+		cellOpen := `<c r="` + cellCoordinate + `" t="` + cellType + `"`
+ 		if styleId > 0 {
+			cellOpen += ` s="` + strconv.Itoa(styleId) + `"`
+ 		}
+		cellOpen += ">" + valueOpen
+		cellClose := valueClose + "</c>"
```
ryho: 👍 This is a lot cleaner to read

file.go
```diff
@@ -249,10 +257,8 @@ func (f *File) MarshallParts() (map[string]string, error) {
 	workbook = f.makeWorkbook()
 	sheetIndex := 1
 
-	if f.styles == nil {
-		f.styles = newXlsxStyleSheet(f.theme)
-	}
-	f.styles.reset()
```
ryho: It worried me that the undocumented logic here was changed, but I can't find any reason why the style was reset when the style sheet was not new, so this is probably safe.

stream_file_builder.go
```diff
@@ -54,6 +55,16 @@ const (
...
 var BuiltStreamFileBuilderError = errors.New("StreamFileBuilder has already been built, functions may no longer be used")
 
+var cellTypeToNumFmtMap = map[CellType]string{
+	CellTypeString: builtInNumFmt[builtInNumFmtIndex_STRING],
+	CellTypeNumeric: builtInNumFmt[builtInNumFmtIndex_INT],
+	CellTypeBool: builtInNumFmt[builtInNumFmtIndex_GENERAL],
+	CellTypeInline: builtInNumFmt[builtInNumFmtIndex_STRING],
+	CellTypeError: builtInNumFmt[builtInNumFmtIndex_GENERAL],
+	CellTypeDate: builtInNumFmt[builtInNumFmtIndex_GENERAL],
+	CellTypeStringFormula: builtInNumFmt[builtInNumFmtIndex_STRING],
```
ryho: Can you document that this is the *default* number format for these types? Previously this library almost entirely conflated the notions of cell type and number format. I've done a lot to distinguish the two concepts in the library, but sometimes it still isn't clear from the code that they are completely different. These are just the default number formats that Excel uses when a cell becomes of a certain cell type.

tealeg:
This was addressed by #442

PR #363, opened 2018-02-13
Fix out of bounds error

ryho:


ryho:
This out of bounds check fix is so trivial, that I am going to merge it without review.

PR #374, opened 2018-04-14
Read & Set Indent

kishorevaishnav:
I think reading the Indent was missing along with other Alignment attributes. Added the code, hope this will fix.


ryho:
Can you add a test that verifies that it fixes it?

tealeg:
I have added testing in support of this and reworked the `getStyles` method.

PR #381, opened 2018-04-22
support excel data validation

rentiansheng:
Data validation support list , displayed in excel drop-down list box

issue  #380


rentiansheng:
@ryho   Hi, Is there any problem with this data validation function implementation?

ryho:
Thanks for this PR, sorry that I don't have enough time to look at this project very often. I am also going to start using this feature in my project.

FYI: you can make AllowBlank, ShowInputMessage, and ShowErrorMessage into Go bools and the XML library will take care of the conversions. I think I'll do that in a follow up PR.

rentiansheng:
I can not set parameters when creating data validation rules.
When SetError sets an error prompt, set the relevant value again.

https://github.com/rentiansheng/xlsx/blob/master/datavalidation.go#L65  cancel parameter.
https://github.com/rentiansheng/xlsx/blob/master/datavalidation.go#L74  set value

PR #385, opened 2018-04-29
New cells are added, no cell format is set.

rentiansheng:
If the cell format is set when the cell is newly added, the cell format set by the column will be lost.

I have added a new SetFormat function to set the cell format. Originally wanted to define the cell format as an enumeration, but too much. Gave up


tealeg:
This PR makes two changes:
- Remove the default numfmt value in NewCell and add a SetFormat method on the cell.
- Add some unrelated time handling functions
The fist of these changes (the one described in the title) has already been made (this PR is now old).  The second seems unrelated and unused.  Therefore I'm closing this PRi

PR #390, opened 2018-05-14
parse XML files about external links.

kawakami-o3:
I'm working on make a formula evaluator.
As a preparation of my job, I have made the patch for parsing XML files about external links.
A file name of external link is included in a formula string. On the other hands, a path to external link
is included in the other XML file. My patch parses this XML file and adds a field "File.ExternalLinks".


tealeg:
Hi @kawakami-o3.  

I'm sorry it took so long for someone to look at this.  If you're still interested this PR would need some tests added before it would be ready for submission.

PR #405, opened 2018-08-06
Fix index out of range.

ricardolonga:


tealeg:
This PR has sat around too long and no longer makes sense.  Sorry about that @ricardolonga.

PR #414, opened 2018-10-05
Add indexed color handling.

akuendig:


tealeg:
@akuendig - nice work, thanks for your efforts.  I've updated this and will now merge.

PR #415, opened 2018-10-08
Add reading hyperlink support, Add 2 font attributes, bug fixes

coraWhite:
Hello! This is my first PR here. I've been using the library daily for reading .xlsx files the past couple months, and have added some read support features, as well encountered a couple of bugs along the way. I understand this is a large PR, but here's a succinct summary of my additions and changes, and I'd be happy to add more detail if requested. I have also browsed the current issues and didn't find my fixes or additions included in any of them to reference. 

* Add read hyperlink support:
  - Add "readSheetRelsFromZipFile()" function for grabbing worksheet relationship files; mimic implementation for grabbing xml worksheets
  - Add "XmlSheetRels" type that holds type "xmlRels", a type to hold Worksheet relationship file attributes, such as hyperlinks, found at "xl/worksheets/_rels" and have the form "sheetN.xml.rels"
  - Add a map "SheetRelationships" as an exported field attribute to File type; Map key is the sheet name "sheetN.xml" to a pointer of a "XmlSheetRels" type
  - Add a test file "xmlSheetRels_test.go" and "testdocs/testHyperlinks.xlsx" to test the validity of grabbing various relationships/hyperlinks in a workbook, across multiple sheets

* Change "Font.Sz" type from Int to Float64: sizes can be floats
  - Update test cases

* Add read Font support for Strikethrough
  - Add Font field attribute "xFont.S" and "style.Font.Strikethrough" with boolean value
  - Add test cases

* Add read Font support for Superscript/Subscript
  - Add Font field attribute "xFont.VertAlign" and "style.Font.Strikethrough" with possible string values of "superscript" or "subscript"  
  - Add test cases

* Add a built in number format to "builtInNumFmt" map
  - With key "5", a common number format code encountered in workbooks

* Bug fixes
  - Alignment.Indent: Change "xf.Alignment.Indent" type from "int" to "string": ooxml treats it as a string, the value wasn't being grabbed at all; update test cases
  - Fill: Swap Fill values "bgColor" and "fgColor" in various calls to "NewFill"; in some instances, new cells were created that had black backgrounds due to the NewFill calls with "000000" as the bgColor; update test cases
  - builInNumFmt: value with key "42" was missing a quotation mark in the string, causing this code to not evaluate correctly
  - addNumFmt(): change logic when determining whether a number format is a built in number format or not; not all number formats are present in the map "builtInNumFmt", so checking if the code is less than the total number is not accurate: changed to query the map instead (line 381 "xmlStyle.go")


GinSanaduki:
→ Add read Font support for Strikethrough

For example, enter TEST as Excel A2. Press ctrl + 5 from the top of the A2 cell.
After saving, if you check the value of Strikethrough in cell.style, it is true.

In contrast, enter TEST2 in Excel B4.
Select only "EST" and press ctrl + 5 to add a strikethrough.
In this case, after saving and checking the Strikethrough value in cell.style, it is false.

Perhaps this will be the same for items owned by other cell.styles.

When VBA implements a search for a character string with a strikethrough line added halfway, there is a memory that was cut out character strings one by one and checked.

If you decide completely, you need to implement such processing.

PR #439, opened 2019-04-02
Improvements to XLSX Data Validation

ryho:
NewXlsxCellDataValidation no longer sets ShowErrorMessage and ShowInputMessage. They are set implicitly by SetInput and SetError.

I added SetInFileList which creates a list validation that references another section in the file instead of a hard coded list of values.

I created constants for the parts of the XLSX cell reference format and used those throughout the code base.

I added a lot of comments from the XLSX spec PDF to the data validation struct.

GetCellIDStringFromCoordsWithFixed allows creating fixed cell references (prefixed with "$").

datavalidation.go and datavalidation_test.go are now data_validation.go and data_validation_test.go, respectively.

I resolved some warnings, there are no longer any error variables named "error" and there are no uncaught errors except those that should not be caught.

cc @rentiansheng


rentiansheng:
you code pretty.  I will pull you code. 

GetCellIDStringFromCoordsWithFixed must be cell in  current sheet . not reference cell  use other sheet

ryho:
It seems to work for the same sheet or a different sheet. Here's an example of an Excel file that I made using this branch of the library. There are drop downs on the first sheet that use the data in the second sheet.
https://1drv.ms/x/s!Ainzlpx-cougghR-gBtkp2-WvYgF

ryho:
@rentiansheng Don't use this code yet. There is a bug in it, so I'm going to make a new PR.

in sheet.go I changed:
```
dd.Sqref = fmt.Sprintf("%s%d:%s%d", colName, dd.minRow, colName, dd.maxRow)
```
to
```
dd.Sqref = colName + RowIndexToString(dd.minRow) + cellRangeChar + colName + RowIndexToString(dd.maxRow)
```
That was a change in behavior. You were using maxRow as the Excel row number, but RowIndexToString expects the number to be zero indexed. This creates an "off by one" error.

rentiansheng:
ok. I know.  I will pay attention later. Here we should specifically define a public method for producing Sqref.    write code in  lib.go

PR #441, opened 2019-04-03
Remove off by one error

ryho:
Also fix a typo


ryho:
@rentiansheng To get rid of the bug I introduced, I made your function accept zero based indexes instead of 1 based indexes.

rentiansheng:
ok

PR #443, opened 2019-04-16
Autofilter

DamianSzkuat:
Hello, 

We would like to create sheets that have  `autoFilter`s on the headers for ease of use in large excel files and as far as we know many people use these filters as well. 

I made a small extension to the API by adding the `AddSheetWithAutoFilters` function. 
This works just like `AddSheet` but will add `autoFilter`s to the header of each column.

To test if the `autoFilter`s are added I also had to add a bit of code in the `readSheetFromFile` function in lib.go that would add the already read `autoFilter`s to the created sheets.


tealeg:
@DamianSzkuat - thanks very much.  Sorry it took so long to get around to this.

Issue #365, opened 2018-02-15
Invalid style number causes Excel 2016 warning


kirkov:
Hi,

I'm using the stream file builder to build the spreadsheet. It works fine on Numbers (Mac) but Excel 2016 (Windows) complains about it the file being broken. I've tracked down the error to be that each cell references an invalid style number.

```
<c r="E2" t="inlineStr" s="2">
...
</c>
```
![image](https://user-images.githubusercontent.com/239877/36270406-eca9cce8-127b-11e8-9a8a-4839389b955c.png)

Each cell references style number 2, but I guess since the `s="NUMBER"` is zero indexed, it should be 1.

If I manually change the values to 1, it works.

This is the line that writes the style number:
https://github.com/tealeg/xlsx/blob/31d1ae382fc11921c2fdc0f8a28f1157b2485b50/stream_file.go#L94

I can do a PR for this but found it odd no one else had this issue.

ryho:
It's not unlikely that this is an issue. I wrote the streaming portion pretty recently and I only use it to force every cell to be a string. I'd be glad to review a PR if you made one.

cauli:
I'm also having the exact same problem. 
Reading the file works fine on Numbers and LibreOffice (Mac), but it complains on Excel (Windows).
The project itself (`tealeg/xlsx`) cannot open the binary with `xlsx.OpenBinary( ... )`, it errors with `index out of range`. 

@kirkov Thanks for looking into this problem. I'll make a PR

pandanrain:
@cauli  thank u. I have tried it works well

alicebob:
Thanks for the suggested fix. It's still `1` on master :(

This is a minimal testcase:

```
package main

import (
    "bytes"

    "github.com/tealeg/xlsx"
)

func main() {
    buf := &bytes.Buffer{}
    f := xlsx.NewStreamFileBuilder(buf)
    ctype := xlsx.CellTypeString
    if err := f.AddSheet(
        "messages",
        []string{"foo", "bar"},
        []*xlsx.CellType{&ctype, &ctype},
    ); err != nil {
        panic(err)
    }

    stream, err := f.Build()
    if err != nil {
        panic(err)
    }
    stream.WriteAll([][]string{
        {"row1a", "row1b"},
    })
    stream.Close()

    if _, err = xlsx.OpenBinary(buf.Bytes()); err != nil {
        panic(err)
    }
}
```

PR #463, opened 2019-07-26
Fix of "Invalid style number causes Excel 2016 warning" #365 

melehin:
A file created by NewStreamFileBuilder opens without recovery messages (in my case via MS One Drive) after this fix.
And this test case works well (without **panic: runtime error: index out of range**)
```
package main

import (
    "bytes"

    "github.com/tealeg/xlsx"
)

func main() {
    buf := &bytes.Buffer{}
    f := xlsx.NewStreamFileBuilder(buf)
    ctype := xlsx.CellTypeString
    if err := f.AddSheet(
        "messages",
        []string{"foo", "bar"},
        []*xlsx.CellType{&ctype, &ctype},
    ); err != nil {
        panic(err)
    }

    stream, err := f.Build()
    if err != nil {
        panic(err)
    }
    stream.WriteAll([][]string{
        {"row1a", "row1b"},
    })
    stream.Close()

    if _, err = xlsx.OpenBinary(buf.Bytes()); err != nil {
        panic(err)
    }
}
```

_Originally posted by @alicebob in https://github.com/tealeg/xlsx/issues/365#issuecomment-506320321_


alicebob:
Great, thanks!

alicebob:
I just tried this branch to generate a document with a few columns and 500K rows, worked great in google docs. Thanks, @melehin!

alicebob:
Hi @ryho, sorry to bother you directly, but this PR fixes the issue I have with producing streaming output. Is there anything you need before you can merge it?

zkanda:
Ping @ryho, would be good if we can merge this.

PR #465, opened 2019-08-05
Added func and procedure to get non-builtin style format

dgrr:
Hello,

I was having an issue with Date formats when the date format wasn't specified in the built-in formats. Diving into your package I realize you weren't using the non-builtin formats. This pull request is intended to add this feature.

Thanks.


dgrr:
What  I should do to recover the coverage? More tests or what?

PR #481, opened 2019-10-15
feat: open with sheet name filter

judwhite:


judwhite:
This definitely was supposed to be in my fork, apologies

PR #523, opened 2020-02-06
added some strings to compare with \ for better compatibility?

acidobinario:
I had some problems opening excel files in windows, after a while of research on how the library works I found that the switch statement on the ReadZipReaderWithRowLimit was comparing the name of the zip files with an '/' char separating the folders, and i had a '\' in the file.Header.Name


tealeg:
Thanks very much - I've never even attempted to run this library on a windows machine in it's more than 10 years of history ;-)

PR #529, opened 2020-03-10
Fix panic when parsing broken hyperlink

gopherclass:
Broken hyperlink is a hyperlink with attribute `r:id` missing. Microsoft Excel opens a file containing broken hyperlink and just leaves it broken. Currently `readSheetFromFile` returns an error if it recognizes a broken hyperlink.

While addressing of the issue, I figured out that `readSheetFromFile` has wrong concurrency implementation, which leads to memory violation panic. I also fixes these too.

According to comment of `readSheetFromFile`, `sheetCount` number of goroutines is created to read each sheet. I think this is not good.


tealeg:
Thanks  - this PR introduced some race conditions because of the change in concurrency model, but I've dealt with those.

gopherclass:
I made assumption on `readSheetFromFile` and commented it. if race condition happens, the assumption is wrong and the comment should be removed.

https://github.com/tealeg/xlsx/blob/f697a605eab6981ea959695096dea33daf341752/lib.go#L656

tealeg:
Thanks for the heads-up, I've removed it.

PR #532, opened 2020-03-24
add support of rich text

kzmi:
This pull request add support of the rich text in the cell.

The points are as follows.

### Read existing XLSX file

`<r>` tags (include children tags like `<rPr>`,`<rFont>`,`<family>`...) in sharedStrings.xml are deserialized to the []RichTextRun.

They can be read from the new field Cell.RichText.

Cell.FormattedValue() returns plain text converted from the rich text.

### Write XLSX file

RefTable can handle both plain text and rich text.

In sharedStrings.xml, []RichTextRun are serialized as `<r>` tags and their children.

To retain white spaces in rich text, `<t xml:space="preserve">` is used to output text fragment which have leading/trailing white spaces.
In addition, conventional normal texts are also processed in the same manner. This may be **a breaking change**.

### Edit Cell

To add rich text into the cell, Cell.SetRichText() is used instead of Cell.SetValue().

[Sample code](https://gist.github.com/kzmi/00b94f597858e4884787e631bc03d56a)

Result

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -60,6 +60,7 @@ func (ct *CellType) fallbackTo(cellData string, fallback CellType) CellType {
 type Cell struct {
 	Row            *Row
 	Value          string
+	RichText       []RichTextRun
 	formula        string
 	style          *Style
 	NumFmt         string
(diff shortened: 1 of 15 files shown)


kzmi:
Ahh...it seems need more work.  I will try to figure out how to do that.

tealeg:
> Ahh...it seems need more work. I will try to figure out how to do that.

If you have any questions feel free to ask.

kzmi:
I added new code to the diskv.go for serialization/deserialization of the new structs.
It might be better to move them to richtext.go to be prepared structural changes, but I didn't do so because tests have to be dependent on DiskVCellStore.

PR #540, opened 2020-04-03
check .Hidden flag for sheet in Marshal

mekegi:
We need render xlsx with hidden sheets. But right now current implementation render always with `State: "visible"`
In current PR - I check Hidden field in Sheet and use it for hiding sheets


svanharmelen:
I have the exact same use case (creating a workbook with a hidden sheet). So 👍 for this one!

PR #564, opened 2020-05-12
simplify ColIndexToLetters()

alicebob:
This is on top of the tag `[v3.0.0]`.

This PR makes `ColIndexToLetters()` a bit simpler. The code comes from another repo, which is credited in a comment, and that repo has an MIT license.

(and I agree, xls column names are weird)

### DIFF
--- a/lib.go
+++ b/lib.go
@@ -72,82 +72,20 @@ func ColLettersToIndex(letters string) int {
 	return sum
 }
 
-// Get the largestDenominator that is a multiple of a basedDenominator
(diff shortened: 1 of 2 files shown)


alicebob:
This is the license, which looks like totally vanilla MIT to me:
https://github.com/psmithuk/xlsx/blob/master/LICENSE

(I don't mind doing it, but that might take a bit)

tealeg:
Cheers @alicebob - all sorted.

alicebob:
Thanks for the quick review!

PR #572, opened 2020-06-01
Emit mergeCells after sheetData for Excel

markwhi:
Microsoft Excel will fail to parse a worksheet with a mergeCells element that occurs before the sheetData element.

The following test code generates the error `Replaced Part: /xl/worksheets/sheet1.xml part with XML error.  Load error. Line 2, column 757.` when attempting to open & repair a workbook generated with the following code:
```go
func main() {
	xls := xlsx.NewFile()
	sht, _ := xls.AddSheet("sheet1")
	row := sht.AddRow()
	cell := row.AddCell()
	cell.Value = "test"
	cell.Merge(1, 0)
	path := "merged-broken.xlsx"
	xls.Save(path)
}
```

### DIFF
--- a/xmlWorksheet.go
+++ b/xmlWorksheet.go
@@ -332,7 +332,7 @@ type xlsxMergeCell struct {
 }
 
 type xlsxMergeCells struct {
-	XMLName  xml.Name                 //`xml:"mergeCells,omitempty"`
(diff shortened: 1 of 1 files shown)


markwhi:
This fix is incomplete, will submit a new PR with the complete fix shortly.

PR #581, opened 2020-06-18
Accept non-pointer values as argument for the Row.WriteSlice method

ninedraft:
There is no need to use pointers to the slices since slices a basically fat pointers by themselfes. 
I propose to permit just plain slices as Row.WriteSlice parameter without breaking API.


write.go
```diff
@@ -7,32 +7,29 @@ import (
 	"time"
 )
 
-// Writes an array to row r. Accepts a pointer to array type 'e',
+// Writes a slice to row r. Accepts slices and pointers to the slices,
```
tealeg, line 10:
```suggestion
// Writes a slice to row r. Accepts a slice or a pointer to a slice,
```

write.go
```diff
@@ -7,32 +7,29 @@ import (
...
 // and writes the number of columns to write, 'cols'. If 'cols' is < 0,
```
tealeg, line 11:
```suggestion
// and will wirte up to the provided number of columns, 'cols'.  If 'cols' is < 0,
```

tealeg: I know you didn't add this comment, but it'd be nice to have it corrected.

ninedraft:
done

tealeg:
@ninedraft thanks very much!

PR #597, opened 2020-07-09
Internalhyperlinks

derlinkshaender:
**Reason for the change**: Currently internal hyperlinks are not supported (they use the attribute "`location`" in the xmlWorkSheet and do not show up in the rels).

**Changes:** I added support for this by appending an exported field `Location` in the Hyperlink struct, added the necessary coe and added a test in `file_test.go` as well as a test doc `testhyperlinks.xlsx`. 
Content of the `location` attribute is either a direct cell address or a named range from the `DefinedNames`.
I'm still learning about the testing framework so I added a separate test function to not break things. 

**Advantages**: allows to process or store internal links in the xlsx file.


cell.go
```diff
@@ -123,10 +123,14 @@ func (c *Cell) key() string {
 
 }
 
+// Hyperlink is a structure to store link information
+// in-workbook links to cells or defined names are stored in Location
+// external links are stores in Link
```
tealeg, line 128: Love that you added documentation comments 👍

file_test.go
```diff
@@ -1040,6 +1040,35 @@ func TestFile(t *testing.T) {
 
 }
 
+// we can unmarshal hyperlinks with in-workbook links
+func TestHyperlinks(t *testing.T) {
```
tealeg, line 1044: Also great that you increased the test coverage.

Issue #602, opened 2020-07-20
How to rename sheet?


Doctor-X:
here is the code that I am using-

`func main() {
	// Read xlsx file
	f, err := xlsx.OpenFile("Schema.xlsx")
	if err != nil {
		log.Println(err)
		os.Exit(-1)
	}
	sh := f.Sheet["Sheet1"]
	sh.Name = "Metadata"
	delete(f.Sheet, "Sheet1")
	f.Sheet["Metadata"] = sh
	f.Save("Schema1.xlsx")

}`

Though "Metadata" sheet is present in Schema1.xlsx, but the content is incorrect.
Any help on this would be highly appreciated. 

Thanks

vladimirfilin:
There’s two bugs here:
1. Rows are missing
2. Issue #595 
![image](https://user-images.githubusercontent.com/60051416/92470752-92bbbe80-f1df-11ea-8ef1-fbcd9e5ff77f.png)

PR #615, opened 2020-08-25
Feature issue remove rename sheet

### LINKED ISSUE #590: 为什么没有删除sheet 的方法？(Why is there no way to delete a sheet?)

rentiansheng:
### feature 

- remove a sheet from file #590
- rename a sheet from file  #602

close #590 #602

### DIFF
--- a/file.go
+++ b/file.go
@@ -450,3 +450,22 @@ func (f *File) RemoveSheet(sheetName string) {
 	f.Sheets = newSheets
 	delete(f.Sheet, sheetName)
 }
+
+// RenameSheet rename  Sheet
+func (f *File) RenameSheet(oleSheetName, newSheetName string) {
+
+	// wait change name sheet
+	var sheet *Sheet
+	for _, sheetItem := range f.Sheets {
+		if sheetItem.Name == oleSheetName {
+			sheetItem.Name = newSheetName
+			break
+		}
+	}
+
+	if sheet != nil {
+		delete(f.Sheet, oleSheetName)
+		f.Sheet[newSheetName] = sheet
+	}
+
+}
--- a/file.go
+++ b/file.go
@@ -436,3 +436,17 @@ func (f *File) ToSliceUnmerged() (output [][][]string, err error) {
 
 	return output, nil
 }
+
+// RemoveSheet remove Sheet from file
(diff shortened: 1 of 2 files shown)


file.go
```diff
@@ -436,3 +436,36 @@ func (f *File) ToSliceUnmerged() (output [][][]string, err error) {
...
+			continue
+		}
+		newSheets = append(newSheets, sheet)
+	}
+	f.Sheets = newSheets
+	delete(f.Sheet, sheetName)
+}
+
+// RenameSheet rename  Sheet
+func (f *File) RenameSheet(oleSheetName, newSheetName string) {
```
Doctor-X, line 455: In RenameSheet(), I can't see where sheet is getting initialized

tealeg, line 455: I believe `oleSheetName` should be `oldSheetName`.

file.go
```diff
@@ -436,3 +436,36 @@ func (f *File) ToSliceUnmerged() (output [][][]string, err error) {
...
+	// wait change name sheet
+	var sheet *Sheet
+	for _, sheetItem := range f.Sheets {
+		if sheetItem.Name == oleSheetName {
+			sheetItem.Name = newSheetName
+			break
+		}
+	}
+
+	if sheet != nil {
```
tealeg, line 466: This block will never run, as `sheet` is never set.

tealeg (changes requested):
It looks good in principle, there's a couple of little tidies needed.

file_test.go
```diff
@@ -1082,3 +1082,36 @@ func (s *SliceReaderSuite) TestFileWithEmptyCols(c *C) {
...
+	c.Assert(f.Sheet["RemoveSheet"], Equals, sheet)
+
+	f.RemoveSheet("RemoveSheet")
+	c.Assert(len(f.Sheets), Equals, 0)
+	var nilSheet *Sheet
+	c.Assert(f.Sheet["RenameSheet1"], Equals, nilSheet)
+}
+
+// Test that we can remove a sheet to a File
+func (l *FileSuite) TesRenameheet(c *C) {
```
Doctor-X, line 1104: This function will never be executed, should be named TestRenameSheet()

tealeg (approved):

Issue #189, opened 2016-01-01
Stretching the column to the width of the widest cell value


orcaman:
Hi,

I'm trying to control the width of a column, ideally what I would be able to do is to stretch the column width to the width of the longest cell value in the column (this is kind of like click the right edge of the top most column in excel). Is this possible to achieve?

Thanks

carstn:
It's not built in to the package yet, but it would be a great feature if you added it! I believe you'll need to manually calculate the width of all glyphs in a given cell, referring to that cell's font. Do that for each cell in a column, and then set the column width.

You could probably find a simpler method to approximate it depending on your needs though.

baconalot:
Related: maybe a quick fix is implementing a bool for "bestFit".
<col min="4" max="4" width="12" bestFit="1" customWidth="1"/>

rbucker:
I'm following up because I wrote a similar issue many months ago and I came to the realization that it's impossible. In fact going back to 1987-ish I remember working on OS/2 and Windows and it was a challenge writing C/C++ code to compute the width. There may be some internals in the Microsoft APIs but that's one challenge... Give this a try and right click on the column header. There is no autoresize function.  Justy something to let you specify the width in a number. So the same is going to apply to your code too.  Personally I take the default font, multiple the number of chars in the formatted cell by 1.6 or 1.8.

Issue #153, opened 2015-08-06
column width fit to size


rbucker:
please add a feature that tell the spreadsheet to set the column width to the size of the data.

tealeg:
@rbucker thanks.  I can't give you an estimate, but we have the request here now and maybe it'll get picked up by someone.  I'd be very happy to see that patch.

rbucker:
I might have to withdraw this one.  According to followup reading it's not possible.  The user needs to compute the adjusted size and then set it manually.  It's tough to do because you have to determine string length in bytes, then the DPI for the selected font, and then consider auto-wrap and manual wrap. My current approximation works but it also sucks. I think we can close this one?

tealeg:
@rbucker I think so too. Thanks anyway!

PR #618, opened 2020-09-08
Autowidth

egpotter:
I tried to implement an autowidth (or autofit) function. There are some issues related to this:

https://github.com/tealeg/xlsx/issues/189
https://github.com/tealeg/xlsx/issues/153

This is not a perfect solution, but I think some people would benefit from it. I separated the proportion function so it can be changed when needed by others. Please let me know in case it does not make sense to merge it, as I need it anyway.

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"strconv"
+	"strings"
 
 	"github.com/shabbyrobe/xmlwriter"
 )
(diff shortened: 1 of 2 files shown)


tealeg:
Thanks for that @egpotter - happy to include it.    Sorry for the delay in reviewing this, I've been working on a major branch and haven't had much time for tending to these things.

egpotter:
Thanks a lot for merging it!

PR #620, opened 2020-09-12
Fixes internal links

Dragomir-Ivanov:
Official format for internal links is in form of `#Sheet!A123`.
Maximum number of hyperlinks per sheet is 65530, according to specification.


tealeg (approved):
Good stuff, thanks for your contribution!

Issue #619, opened 2020-09-08
Using sheet.Row() kills the first row of the xlsx file


vladimirfilin:
Accidentally discover when digging in #617 . 
If we use sheet.Addrow(), the resulting file is correct:
````go
package main

import (
	"log"

	"github.com/tealeg/xlsx"
)

func main() {
	var (
		sheet *xlsx.Sheet
		row   *xlsx.Row
		cell  *xlsx.Cell
		err   error
	)
	f := xlsx.NewFile()
	sheet, err = f.AddSheet("Data")
	if err != nil {
		log.Fatal(err)
	}

	// AddRow() part
	row = sheet.AddRow() // 1
	cell = row.AddCell() // A
	cell.SetString("1 Val:")
	cell = row.AddCell() // B
	cell.SetInt(1)
	log.Printf("Row added, id: %v, MaxRow: %v", row.GetCoordinate(), sheet.MaxRow)

	row = sheet.AddRow() // 2
	cell = row.AddCell() // A
	cell.SetString("2 Val:")
	cell = row.AddCell() // B
	cell.SetInt(2)
	log.Printf("Row added, id: %v, MaxRow: %v", row.GetCoordinate(), sheet.MaxRow)

	err = f.Save("result1.xlsx")
	if err != nil {
		log.Fatal(err)
	}
}
````
log:
````
2020/09/08 18:36:57 Row added, id: 0, MaxRow: 1
2020/09/08 18:36:57 Row added, id: 1, MaxRow: 2
````
[result1.xlsx](https://github.com/tealeg/xlsx/files/5189545/result1.xlsx) has two rows:
![image](https://user-images.githubusercontent.com/60051416/92501766-a8dd7500-f207-11ea-8809-73958aea2699.png)

Let’s add a couple of rows using sheet.Row():
````go
package main

import (
	"log"

	"github.com/tealeg/xlsx"
)

func main() {
	var (
		sheet *xlsx.Sheet
		row   *xlsx.Row
		cell  *xlsx.Cell
		err   error
	)
	f := xlsx.NewFile()
	sheet, err = f.AddSheet("Data")
	if err != nil {
		log.Fatal(err)
	}

	// AddRow() part
	row = sheet.AddRow() // 1
	cell = row.AddCell() // A
	cell.SetString("1 Val:")
	cell = row.AddCell() // B
	cell.SetInt(1)
[...]

PR #621, opened 2020-09-13
fix: Using sheet.Row() kills the first row of the xlsx file #619

shawn42025:


tealeg:
Hi @shawn42025  - I think the master branch no longer suffers from this issue, owing to the major branch I just committed.  I'll leave this PR open for now, just in case it's still an issue.   Sorry for being a bit slow on this, I've had my hands full.

Issue #641, opened 2020-11-10
DefaultRowHeight is missing v3.2.1


stvoidit:
after running the tests, when I got rid of everything that caused the panic, I was a little surprised, where is all the data in the file?

@tealeg 
I think you need to leave this parameter:
**worksheet.SheetFormatPr. DefaultRowHeight = 12.85**

Otherwise, all lines are "glued together". I don't think you should leave the cell height 0 at all.
Either I don't quite understand the concept of upcoming changes yet, sorry :) @tealeg

tealeg:
@stvoidit oops ;-) I'll look into it.

tealeg:
Can you try with this branch and let me know if this fixes your issue? https://github.com/tealeg/xlsx/tree/force-default-row-height-above-zero

stvoidit:
> 
> 
> Can you try with this branch and let me know if this fixes your issue? https://github.com/tealeg/xlsx/tree/force-default-row-height-above-zero

@tealeg 
hmmm... I think something went very wrong :D
Example of my tests in the archive.

Setting the default row height didn't work for some reason.
I also don't really understand how the coordinates "floated" during rendering.

I think I have a simple enough example for you to track this behavior.
[v3xlx.zip](https://github.com/tealeg/xlsx/files/5522725/v3xlx.zip)

![Screenshot 2020-11-11 111105](https://user-images.githubusercontent.com/43279538/98786329-02803c80-240f-11eb-9ef2-a09eea190042.png)

tealeg:
@stvoidit thanks.

PR #642, opened 2020-11-14
Address issues with the value of row.maxCol and sheet.maxCol

tealeg:
This PR addresses mutliple issues resulting from the incorrect management of the row.maxCol and Sheet.MaxCol values in both the MemoryCellStore and the DiskVCellStore.


tealeg:
Fixes #641

PR #691, opened 2021-03-30
Fix empty cells in Mac OS Numbers

### LINKED ISSUE #594: Empty cells are filled by the first column name - v3.2.0
Hi there!

After an update version 2.x to 3.2.0 during our tests, we realized that now the cells that were empty are being filed with the name of the first column. 
By the way, this behavior is just happening on Apple Numbers. I've checked on Microsoft Excel and Google Docs and it worked well on both.

**Expected behavior**
correct_xls_on_apple_numbers

**Unexpected behavior:**
error_xlsx_apple_numbers

donners:
When no string value is set for a cell, Numbers copies the value from cell A1. Setting an empty string fixes the issue.

This fixes #594. I did see that the release notes for v3.2.1 mentioned this being fixed already, but I'm still seeing the Numbers issue on current master branch.


tealeg (approved):
Thanks @donners

Issue #687, opened 2021-03-12
The name of second worksheet of `tutorial/samplefile.xlsx` is `Styles` but not `Cities`


northbright:
Hi Teale

The second worksheet name seems to be `Styles` when I download the `tutorial/samplefile.xlsx` and run the example code:
```
package main

import (
        "fmt"
        "github.com/tealeg/xlsx/v3"
)

func main() {
        wb, err := xlsx.OpenFile("./samplefile.xlsx")
        if err != nil {
                fmt.Printf("Open XLSX file error: %v\n", err)
                return
        }

        for i, sh := range wb.Sheets {
                fmt.Println(i, sh.Name)
        }
}
// Output:
// 0 Sample
// 1 Styles
```

I found it should be `Cities` in the tutorial:
> The second worksheet is named Cities and contains a list of the largest cities of the world

Thanks

PR #692, opened 2021-04-03
Renamed second worksheet from 'Styles' to 'Cities'

rstefanic:
I was also going through the tutorial and noticed this. Small change, but I just renamed the worksheet on the second page from "Sample" to "Cities".

Using the same code snippet from #687 with this fix, I got the following output:

    0 Sample
    1 Cities


tealeg:
Thanks @rstefanic

Issue #647, opened 2020-11-17
Hidden rows in new document with v3.2.x


kpym:
When I :
1) create a new document with `doc := xlsx.NewFile()`, 
2) add a sheet with `sh, err := doc.AddSheet("some name")`
3) add rows with `row := sh.AddRow()` 
4) add cells with `cell := row.AddCell()`
5) and finally save with `doc.Save("filename.xslx")`

my rows are hidden (even if I add `row.Hidden = false` and `cell.Hidden = false`) in version `v3.2.2` and `v3.2.1`.
But they are visible in `v3.1.1`.

mmargerumImpactrx:
Is there a workaround for this until the new release?

wenj1104:
I have the same problem.

baloo32:
This is related to issue #654 I believe... 3.2.0 is last know working for me.

gourlaa:
Same for us

Rocksus:
A simple workaround for me (tested in v 3.2.3) was to add row.SetHeight(12.5) or any pts that you prefer for each row.

Here's an example code:

```Go
package main

import (
	"fmt"

	xlsx "github.com/tealeg/xlsx/v3"
)

func main() {
	fmt.Println("creating document")
	doc := xlsx.NewFile()
	sh, _ := doc.AddSheet("Sheet 1")
	for i := int(0); i < 10; i = i + 1 {
		row := sh.AddRow()
		cell := row.AddCell()
		cell.SetString(fmt.Sprintf("row %d", i))
		row.SetHeight(12.5)
	}
	doc.Save("test.xlsx")
}
```

margerum:
I can't believe this still isn't fixed.  Gonna have to move to something else.

njern:
Noticed this same problem, can confirm the suggested workaround solved it for me as well.

amincheloh:
This problem also occurs in file from `xmlx.OpenFile`.

After use git bisect I found this problem first appears from https://github.com/tealeg/xlsx/commit/995e27c23eed7304116efd3ef762ae0897086536 commit.

PR #694, opened 2021-04-06
fix(647): custom row height checker

chris-buesser:
- Rows were being marked for a custom height attribute of 0 if you added
  a cell. This was hiding the rows (by default) in excel.
- With new customHeight checker, you only attribute "customHeight" to a
  row which has its height set (you can still set this to 0)

ISSUE: tealeg/xlsx #647


tealeg:
@chris-buesser thanks very much for you work!

Issue #689, opened 2021-03-24
UseDiskVCellStore is creating too many files


TobiasSoares:
Hi

I have a problem when opening big files. I've read the documentation and saw that in order to avoid lots of memory allocation I should use the flag `UseDiskVCellStore` in the `xlsx.OpenFile()` statement. 

But when I do that, the function creates at least 1 file per cell, and for big xlsx files that could mean more than 500.000 files (100.000 rows x 5 columns) which is too much for the inodes in the filesystem. I get  the error `no space left in the device` because all the inodes are used up.

If I have to use smaller files to use DiskV then it defeats the purpose of using it. How can I correctly use this feature?

I'm using Fedora 33, don't know if it matters.

tealeg:
The answer to this would be to use a different storage library than DiskV and implement a new cell store.  It is designed to be extensible for because every storage solution provides different compromises (some are slower, some require using C libraries, etc. etc..)  DiskV is provided "in the box" because its pure Go and introduces very little overhead to the library, but you could implement one with pretty much any database.

Alternatively, depending on your use case you could evaluate completely different solutions.  This library is, really not well architected for this kind of work - it evolved organically from a simple hack for reading these files into something with way more capabilities - but really it should have been rewritten to do that.  The CellStores are a way to try and make this library cope a little better, but I'm not surprised you're reaching limitations in its design.   If your task is simply to read data from an Excel file, I suggest having a look at how this library looked circa the [go1.1 tag](https://github.com/tealeg/xlsx/tree/go1.1)  - it's probably still not that memory efficient, but it would be easier to see how reading the Excel files works that the current state.    Otherwise, perhaps one of the other Excel libraries available for go these days could help?

PR #695, opened 2021-04-06
fix: testing temp files cleanup

chris-buesser:
- When testing, files are piling up quickly in the temp directory, now
  files are cleaned up after each run (if the test has not failed).

ISSUE: NONE


chris-buesser:
Somewhat related to #689

tealeg (approved):
Thanks @chris-buesser

PR #701, opened 2021-05-08
add ValueOnly option for FileToSlice

### LINKED ISSUE #700: we use this xlsx.FileToSlice method to decode .xlsx file, but it consumes a lot of memory
Hi, We call the ```xlsx.FileToSlice``` method to parse files uploaded by customers

 <https://github.com/erda-project/erda/blob/develop/pkg/excel/decode.go> 
```
func Decode(r io.Reader) ([][][]string, error) {
	tmpF, err := ioutil.TempFile("", "excel-")
	if err != nil {
		return nil, err
	}
	if _, err := io.Copy(tmpF, r); err != nil {
		return nil, err
	}
	data, err := xlsx.FileToSlice(tmpF.Name())
	if err != nil {
		return nil, err
	}
	return data, nil
}
```

It will run normally until someone uploads an xlsx file with many null values, and these values ​​are not needed.

We found that the xlsx file is only 2M, but this method requires nearly 2G of memory to complete the decoding.

We unzipped the file and found that xl/worksheets/sheet1.xml has 757,875 lines, and most of them are null values. like this
```
<c r="H68" s="50"/>
<c r="I68" s="50"/>
``` 
These null values ​​are distributed in different rows and different columns. Therefore we feel that decoding this unusual xml file consumes a lot of memory. I was wondering if we have some way to prevent this from happening

Below is our screenshot
image
image

QvodSoldier:
This PR fix #700 

if ValueOnly is true, all null-valued cells will be deleted, before decode worksheet.xm. this option can save memory and time when parsing files with a large number of NULL values. But it may also cause accidental injury, because NULL maybe not really meaningless, it also could means empty or somethings else in sheet.


file.go
```diff
@@ -41,6 +42,16 @@ func RowLimit(n int) FileOption {
 	}
 }
 
+// ValueOnly treats all NULL values as meaningless and it will delete all NULL value cells,
+// before decode worksheet.xml. this option can save memory and time when parsing files
+// with a large number of NULL values. But it may also cause accidental injury,
+// because NULL maybe not really meaningless, it also could means empty or somethings else in the sheet.
```
tealeg, line 48:
```suggestion
// because NULL may not really be meaningless.  Use with caution. 
```

lib.go
```diff
@@ -1126,3 +1128,83 @@ func truncateSheetXML(r io.Reader, rowLimit int) (io.Reader, error) {
 	}
 	return output, nil
 }
+
+// truncateSheetXMLValueOnly will take in a reader to an XML sheet file and will return a reader that will read an equivalent
+// XML sheet file without null vaules of rows. This greatly speeds up XML unmarshalling when we
+// only need non-NULL data for the sheet.
+// When sheets are truncated, most of formatting present will be not right, but all of this formatting
+// is related to printing and visibility, and is out of scope for most purposes of this library.
+func truncateSheetXMLValueOnly(r io.Reader) (io.Reader, error) {
```
tealeg, line 1137: Though I understand the caveats around the meaning of NULL, I really like your approach to this.  Preprocessing the XML whilst maintaining the Reader interface is a smart hack.  👍🏻

tealeg (changes requested):
👍🏻  This is good - I appreciate you pointing out that this is a "risky" option, but I think that's fine as an option (that's why they're options!).  

I've made a suggestion above to improve the English in one of the doc comments, other than that I'm happy.

file.go
QvodSoldier: Thanks, I am also working hard to improve my English 😂. I have corrected it.

tealeg: Great, thanks.  I could have just corrected it myself, but it's always good to share.  As a native English speaker living in Germany, I'm very conscious that one can't improve in a second language if nobody ever corrects you.    I'll go ahead and merge this now.  Thanks again.

tealeg (approved):

tealeg:
I merged this manually after fixing some tests.

PR #720, opened 2021-07-08
fix datavalidation dorplist size

yixinin:
ms office allow max 255 runes in datavalidation input, not 255 bytes.

### DIFF
--- a/data_validation.go
+++ b/data_validation.go
@@ -3,6 +3,7 @@ package xlsx
 import (
 	"fmt"
 	"strings"
+	"unicode/utf8"
 )
 
 type DataValidationType int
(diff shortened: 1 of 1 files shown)


yixinin:
pls also fix v1 and v2 tag, thx!!!

rentiansheng:
Is this an error caused by the encoding problem causing the character length to be different? I don't understand coding. 257 is the limit of ms office excel, if it exceeds the excel opening, it will need to be repaired

yixinin:
实际上excel (office 2016)的数据校验是支持256个字符的，亲测，256个中文字符~~

rentiansheng:
> 实际上excel (office 2016)的数据校验是支持256个字符的，亲测，256个中文字符~~

This is the problem of character set and encoding. I don’t know if there is a compatibility problem, such as across systems and machines.  @yixinin

PR #730, opened 2021-09-11
rename function for lonely style and linters check

geoirb:


geoirb:
success build needs go > 1.15

tealeg:
Sorry, but no.

Many of these corrections are indeed valid and useful, but others are just plain wrong (XLSX is the correct capitalisation, not Xlsx) and worse some of them change the public interface of the package, which would break downstream projects.   I am not actively maintaining this code, and I have no intention of sorting through the output of an automated lint.  Sorry.

PR #735, opened 2021-12-12
colLimit function

mauserzjeh:
Large files can still cause a problem when you want to read all the lines but not all columns. With colLimit now there is a possibility to limit the reading to the set number of columns. Works similiarly to rowLimit, it will only read the amount of columns from the beginning.


tealeg (approved):
+1

PR #744, opened 2022-06-08
fix: memory leak risk when file write err

echoface:
Here a risk memory leak without closing file descriptor correctly, when disk write fail.


tealeg (approved):
Great work, good catch!

PR #750, opened 2022-11-29
Add support for StreamCell merging and custom row heights

colinmroberts:
Adds support for `Merge()` and `Height()` for `StreamCell`/`StreamRow`.

To merge a StreamCell, call `Merge(h, v)` with the number of extra cells to merge in each direction. Note: does not do any advanced bounds/overlap checking, just ensures that it doesn't go beyond the width of the stream sheet.

To set a custom height, create a `StreamRow` and call `Height()`. It must be written to the sheet with `WriteRowS()`.

This PR also resolves a few tests that were broken on the tip of the v1 branch related to styles and date formatting.

### DIFF
--- a/xmlWorksheet.go
+++ b/xmlWorksheet.go
@@ -302,7 +302,7 @@ type xlsxMergeCell struct {
 }
 
 type xlsxMergeCells struct {
-	XMLName xml.Name        //`xml:"mergeCells,omitempty"`
+	XMLName xml.Name        `xml:"mergeCells,omitempty"`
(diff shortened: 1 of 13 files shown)


xmlWorksheet.go
```diff
@@ -302,7 +302,7 @@ type xlsxMergeCell struct {
 }
 
 type xlsxMergeCells struct {
-	XMLName xml.Name        //`xml:"mergeCells,omitempty"`
```
ryho, line 305: Weird.

ryho (approved):

PR #758, opened 2023-05-11
go.mod: upgrade Go to 1.18 to split dependencies

dolmen:
Upgrade "go" in go.mod to 1.18 for dependencies references cleanup (split direct vs indirect).

```console
$ go mod edit -go=1.18 && go mod tidy
```

I think that this cleanup step is needed before upgrading dependencies.


dolmen:
:tada:

Thanks @tealeg. Could you have a look at #759 which is the follow-up?

PR #763, opened 2023-05-23
README: update badges, link to pkg.go.dev

dolmen:
README.org:
* update badges
* link to pkg.go.dev instead of godoc.org


tealeg (approved):
Thanks, looks good.

Issue #747, opened 2022-06-28
Resource leakage in the lib.go file


zhangyue9999:
1、rc, err := worksheetRelsFile.Open() in the makeHyperlinkTable function does not close the RC handle. As a result, resource leakage occurs.
2、In the readSheetsFromZipFile function, rc, err = f.Open() does not close the rc handle, causing resource leakage.
3、In the readSharedStringsFromZipFile function, rc, err = f.Open() does not close the rc handle, causing resource leakage.
4、In the readStylesFromZipFile function, rc, err = f.Open() does not close the rc handle, causing resource leakage.
5、In the readThemeFromZipFile function, rc, err = f.Open() does not close the rc handle, causing resource leakage.
6、In the readWorkbookRelationsFromZipFile function, rc, err = workbookRels.Open() does not close the rc handle, causing resource leakage.

dolmen:
@zhangyue9999 Thanks for this report. I'm working on a fix.

PR #766, opened 2023-06-07
Fix read zip leaks

dolmen:
* Fix leaks in lib.go reported by @zhangyue9999 in #747
* Fix leak in OpenFile


tealeg (approved):
LGTM

tealeg:
Thanks @dolmen

PR #767, opened 2023-06-07
Enable parallel testing

dolmen:
Enable parallel testing to reduce test clock time.

Parallelism is not very aggressive for now because I found an issue with `quicktest` handling of Parallel.

```
go test -parallel=10
```


tealeg (approved):
Nice, really like these small fixes with impact.

PR #770, opened 2023-07-17
WriteStruct and WriteSlice no longer ignore sql.NullTime values

enijns:


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/tealeg/xlsx/v3
+module github.com/enijns/xlsx/v3
```
tealeg, line 1: This change would break the go.mod file.

tealeg (changes requested):
It looks good except for the go.mod change.

go.mod
enijns: sorry about that, this is now fixed in the latest commit

tealeg (approved):

PR #773, opened 2023-09-09
use random file name for RoundTripFileWithNoSheetCols test

benedictjohannes:
In my test, RoundTripFileWithNoSheetCols  fails with error `e"OpenFile: open testdocs/after_write.xlsx: no such file or directory"` This is most likely due to the parallel test runs called by csRunO that try to write and remove the files concurrently. [Failed error attached](https://github.com/tealeg/xlsx/files/12566807/failedTest.txt)

By choosing a random filename,
each test create/write and removes its own file,
allowing this test to pass as expected.

### DIFF
--- a/lib_test.go
+++ b/lib_test.go
@@ -3,7 +3,9 @@ package xlsx
 import (
 	"bytes"
 	"encoding/xml"
+	"math/rand"
 	"os"
+	"strconv"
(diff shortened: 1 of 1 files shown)


tealeg (approved):
Thanks!

PR #774, opened 2023-09-09
don't write dump.xml on truncateSheetXMLValueOnly

benedictjohannes:
It looks like `dump.xml` are merely dumping XML rows that are never read back.

As such it is unnecessary and might cause unexpected errors when the program is called in a read-only directory/FS without serving any purpose.


tealeg (approved):
Agreed - I suspect it was used for debugging and got forgotten.

approve.

PR #779, opened 2023-09-13
formatting correction for README.org

benedictjohannes:
@tealeg you seem to use `md`, while the README is in `org`

Would you mind if I reformat both `README.org` and `tutorial/tutorial.adoc` to markdown? No change will be made to the content. 
Afterwards I will open another PR.


tealeg (approved):
Whoops..  good catch.

PR #780, opened 2023-09-13
allow-subsecond-settime

benedictjohannes:
Currently, `cellRef.SetDateWithOptions` will truncate the supplied `time.Time` to **second** accuracy.

This allows fractional/subsecond part to be passed to Excel.


tealeg (approved):
Very good 👍🏻

PR #781, opened 2023-09-13
allow recognition of time format in multisection numFormat

benedictjohannes:
In [`format_code.go:248-254`](https://github.com/tealeg/xlsx/blob/ac8326d/format_code.go#L248-L254), it was documented that 

> Time formats cannot have multiple groups separated by semicolons, there is only one format.
> Strings are unaffected by the time format.

Apparently Excel 2021 has a *default* numFormat option that violate that condition: `m/d/yy\ h:mm;@`  
It has two formatting sections. Here is a [video](https://github.com/tealeg/xlsx/assets/25011791/79877497-d2a2-46fc-a54a-77af774212be) I recorded in creating this [Excel file](https://github.com/tealeg/xlsx/files/12601286/wb.xlsx).

This commit/PR allows recognition of time format when more than one numFormat section is present, with
- first / positive section having `isTimeFormat` as identifier for `isTimeFormat` for the parsed format string
- formatting for time formatted cell is taken from first/positive section's full format string.

Btw, I am coding an Excel parser that needs to `cellRef.IsTime()` to accurately identify date/time cells. This was the original issue that propelled me to observe and contribute.

### DIFF
--- a/format_code.go
+++ b/format_code.go
@@ -245,13 +245,6 @@ func parseFullNumberFormatString(numFmt string) *parsedNumberFormat {
 	parsedNumFmt := &parsedNumberFormat{
 		numFmt: numFmt,
 	}
-	if isTimeFormat(numFmt) {
-		// Time formats cannot have multiple groups separated by semicolons, there is only one format.
-		// Strings are unaffected by the time format.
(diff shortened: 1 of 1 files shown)


tealeg (approved):
Looks good.

PR #782, opened 2023-09-25
Support bool type values on Cell.SetValue()

WKBae:
`Cell.SetValue(bool)` sets the cell's value to be a plain string "true"/"false", instead of the expected boolean values, displayed as "TRUE" or "FALSE" on MS Excel.
Add the missing bool type case in SetValue to call `Cell.SetBool`.


tealeg (approved):

tealeg:
Thanks @WKBae !

Issue #727, opened 2021-08-03
Can not open excel files when added AutoFilter


linzhicheng5512:
1.Version: v3.2.3
2.Test code:
```
import (
	"github.com/tealeg/xlsx/v3"
)

type structTest struct {
	IntVal     int    `xlsx:"0"`
	StringVal  string  `xlsx:"1"`
	FloatVal   float64 `xlsx:"2"`
	IgnoredVal int     `xlsx:"3"`
	BoolVal    bool    `xlsx:"4"`
}

func main() {
	structVal := &structTest{
		IntVal:     16,
		StringVal:  "hello world",
		FloatVal:   3.14159216,
		IgnoredVal: 7,
		BoolVal:    true,
	}

	//create a new xlsx file and write a struct
	file := xlsx.NewFile()
	sheet, _ := file.AddSheet("TestSheet")
	row := sheet.AddRow()
	row.SetHeight(12.5)
	row.WriteStruct(structVal, -1)
	sheet.AutoFilter = &xlsx.AutoFilter{TopLeftCell: "A1", BottomRightCell: "E1"}
	err := file.Save("test.xlsx")
	if err != nil {
		panic(err)
	}
}
```
3.Open 'test.xlsx' file, have this error message
![image](https://user-images.githubusercontent.com/46001553/127988080-cd5af4ce-9a58-4c04-b1d2-e96d4b3b7181.png)
4.This issue is similar to the [https://github.com/tealeg/xlsx/issues/679](url)

Issue #424, opened 2018-11-22
AutoFilter - Excel crash


mickesommar:
Then adding AutoFilter, excel crashes then sorting. But if I use the filter version first, then sorting is working.
Have tried with Excel 2010, 2013, Windows 7, Windows 10, Windows Server 2012.
In LibreOffice, the autofilter will not a peer.
Compiled with Go 1.11.1

The below code will create the file: names.xlsx, with AutoFilter. 
Open the file, try to sort in the "Name" column. Excel will crash.
But if you try to use the filter, for example, only show the name "bob", then select all and sort. It will work.

Do I use the package wrong?

`
package main

import (
	"github.com/tealeg/xlsx"
)

func main() {
	var file *xlsx.File
	var sheet *xlsx.Sheet
	var row *xlsx.Row
	var cell *xlsx.Cell
	var err error

	file = xlsx.NewFile()
	sheet, err = file.AddSheet("Sheet1")
	if err != nil {
		panic(err)
	}

	// Header
	row = sheet.AddRow()
	cell = row.AddCell()

	// Name 1
	cell.Value = "Name"
	row = sheet.AddRow()
	cell = row.AddCell()
	cell.Value = "Mike"

	// Name 2
	row = sheet.AddRow()
	cell = row.AddCell()
	cell.Value = "Bob"

	// Name 3
	row = sheet.AddRow()
	cell = row.AddCell()
	cell.Value = "Kent"

	// Add autofilter.
	sheet.AutoFilter = &xlsx.AutoFilter{
		TopLeftCell:     "A1",
		BottomRightCell: "A4",
	}

	// Save file.
	err = file.Save("names.xlsx")
	if err != nil {
		panic(err)
	}
}
`

korpa:
Any news on this? I have the same issue.

tealeg:
@korpa sorry, no, I haven't even begun to look into this.

korpa:
I quickly checked the sheet1.xml. The only difference I saw between AutoFilter added by Excel and via the lib is that the position of the AutoFilter Node is different:


AutoFilter added via Excel:
![image](https://user-images.githubusercontent.com/2982785/109963743-0b15bf00-7ced-11eb-8fe6-6a73da427dc5.png)

AutoFilter added via lib:
![image](https://user-images.githubusercontent.com/2982785/109963928-44e6c580-7ced-11eb-8688-2fbf91acdfa2.png)

korpa:
[...]

PR #792, opened 2024-01-03
Fix Excel opening files with AutoFilter data

neddy299:
Microsoft Excel is unable to open xlsx generated Excel files that contain AutoFilter data. This fix saves the autoFilter XML tag after sheetData and the Excel files work as expected.

Addresses the following issues
- https://github.com/tealeg/xlsx/issues/424
- https://github.com/tealeg/xlsx/issues/727


tealeg (approved):
👍🏻

tealeg:
Thanks @neddy299

Issue #793, opened 2024-01-25
It appears that Cell.GetTime() is returning an incorrect time.


martianwei:
Below is in screenshot of my xlsx file.

![螢幕擷取畫面 2024-01-25 133905](https://github.com/tealeg/xlsx/assets/131415064/9cee96a6-8de2-4bb6-ba5b-6364fe629ab9)

I attempted to use Cell.GetTime() to retrieve the time value from my cell, but it seems to be producing an erroneous result.

![螢幕擷取畫面 2024-01-25 134105](https://github.com/tealeg/xlsx/assets/131415064/30b6003f-501a-4f71-9e2c-a44d67a8b168)
![螢幕擷取畫面 2024-01-25 134400](https://github.com/tealeg/xlsx/assets/131415064/c2a483f3-5197-4830-9d71-a96cd582ffd9)

Could you help me fix it?

[ISO_12hours.xlsx](https://github.com/tealeg/xlsx/files/14047341/ISO_12hours.xlsx)

PR #795, opened 2024-01-26
fix: TimeFromExcelTime

martianwei:
issue #793 
fix TimeFromExcelTime


tealeg:
It seems this change causes existing tests to fail.   This might well be more complicated than adding that offset.

tealeg (changes requested):
I think this needs a more thorough investigation and explanation.

Issue #794, opened 2024-01-25
addNumfmt() in xmlStyle.go seems to be wrong


martianwei:
The addNumfmt function verifies whether xNumFmt.NumFmtId is less than or equal to builtinNumFmtsCount. If this condition is met, the function will not add it.

![螢幕擷取畫面 2024-01-25 135947](https://github.com/tealeg/xlsx/assets/131415064/41e203a9-043c-4d96-bcfc-6a2e2c0490f2)

However, since my numFmtId is 8, this function prevents me from retrieving the correct NumFmt for my cell.

![螢幕擷取畫面 2024-01-25 140625](https://github.com/tealeg/xlsx/assets/131415064/7c577cad-ef11-4b06-b47c-7e74516057ec)

I believe it would be more effective to check if xNumFmt.NumFmtId is included in builtInNumFmt

![螢幕擷取畫面 2024-01-25 140913](https://github.com/tealeg/xlsx/assets/131415064/b68c5994-823c-4446-ac48-5150ab7b1c88)

Could Anyone help me fix it?

[Number_Currency.xlsx](https://github.com/tealeg/xlsx/files/14047419/Number_Currency.xlsx)

tealeg:
Fixed by #796.

PR #796, opened 2024-01-26
fix: addNumfmt

martianwei:
issue #794 
fix addNumfmt


tealeg:
@martianwei thanks for this!

PR #798, opened 2024-02-12
fix AutoFilter not working in LibreOffice

martoche:
AutoFilter doesn't work in LibreOffice unless a special `FilterDatabase` tag is present in the `DefinedNames` array.  See:

- https://github.com/SheetJS/sheetjs/issues/1165
- https://bugs.documentfoundation.org/show_bug.cgi?id=118592

This pull request adds the workaround that makes AutoFilter work in LibreOffice.


tealeg (approved):
Thanks @martoche

PR #816, opened 2024-09-12
fix: wrong cell & border colors using indexedColors

### LINKED ISSUE #815: style error
code
```
	v, _ := xlsx.OpenFile("d.xlsx")
	v.Save("e.xlsx")
```
`d.xlsx`  style 
image
after save to `e.xlsx` and open it
image
the style change, the header color change to black

the origin `d.xlsx` file 
[d.xlsx](https://github.com/user-attachments/files/16815023/d.xlsx)

beta:
Closes #815.

- Indexes of indexed colors should be zero-based.
- Indexed colors used in border styles are missing from output.

Tested with MS Excel and Numbers in macOS. Need more testing with other OS's and spreadsheet software.

### DIFF
--- a/xmlStyle.go
+++ b/xmlStyle.go
@@ -237,16 +237,15 @@ func (styles *xlsxStyleSheet) populateStyleFromXf(style *Style, xf xlsxXf) {
 	style.ApplyAlignment = xf.ApplyAlignment
 
 	if xf.BorderId > -1 && xf.BorderId < styles.Borders.Count {
-		var border xlsxBorder
-		border = styles.Borders.Border[xf.BorderId]
+		border := styles.Borders.Border[xf.BorderId]
 		style.Border.Left = border.Left.Style
-		style.Border.LeftColor = border.Left.Color.RGB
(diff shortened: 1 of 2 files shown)


beta:
Original file from #815: [d.xlsx](https://github.com/user-attachments/files/16977916/d.xlsx)

image

Wrong output: [e.xlsx](https://github.com/user-attachments/files/16977950/e.xlsx)

image

Fixed output: [e.xlsx](https://github.com/user-attachments/files/16977944/e.xlsx)

image

tealeg:
Yes, this is the issue. Thanks very much for the PR.   I am working on changing the way colors work for a next major version so that we can round-trip indexed colors, for example, instead of converting them to RGB values, but this fix will do fine for the v3 series.

tealeg:
There's a failing test in the MemoryCellStore right now.  I'll take a look.

tealeg:
Yeah - there was a hidden bug in the implementation of IndexedColors on the cell.  It wasn't caused by your change, just hidden by having the wrong behaviour previously.  All good.  Thanks for your work @beta !

lpxxn:
👍

PR #818, opened 2024-09-12
issue 817 autoFilterBounds could have one element only, in this case …

ivantextmagic:
…use it as second param of AutoFilter

### DIFF
--- a/sheet.go
+++ b/sheet.go
@@ -951,7 +951,7 @@ func handleNumFmtIdForXLSX(NumFmtId int, styles *xlsxStyleSheet) (XfId int) {
 
 func IsSaneSheetName(sheetName string) error {
 	runeLength := utf8.RuneCountInString(sheetName)
-	if runeLength > 43 || runeLength == 0 {
+	if runeLength > 31 || runeLength == 0 {
 		return fmt.Errorf("sheet name must be 31 or fewer characters long.  It is currently '%d' characters long", runeLength)
 	}
 	// Iterate over the runes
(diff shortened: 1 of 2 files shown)


sheet.go
```diff
@@ -949,10 +949,9 @@ func handleNumFmtIdForXLSX(NumFmtId int, styles *xlsxStyleSheet) (XfId int) {
 	return
 }
 
-
 func IsSaneSheetName(sheetName string) error {
 	runeLength := utf8.RuneCountInString(sheetName)
-	if runeLength > 31 || runeLength == 0 {
```
tealeg, line 955: This doesn't seem to be related to the issue you described?  Why do we need to change this?

ivantextmagic: sorry, yes. but I also, have got this problem on this file, please try it out too
[auto-filter-titles.xlsx](https://github.com/user-attachments/files/16993125/auto-filter-titles.xlsx)

tealeg: @ivantextmagic - to my knowledge Excel still enforces 31 charaters as a limit, and historically it certainly did.  Whilst the file format doesn't enforce that limit, and LibreOffice will open it just fine, it's generally not wise to create XLSX files that don't work in Excel.  In general Excel is the *most* picky about XLSX files. 

We could introduce a separate FileOption to support sheet names longer than 31 characters, but that would need to be a separate PR please.

ivantextmagic: okay,returned it to 31
but still, I receive some files with characters number > 31
I'm not sure about the source, but I think it could be some old business systems which could export their data to xlsx

PR #820, opened 2024-09-19
fix: new empty row index bug

lpxxn:
This PR will introduce a bug.
https://github.com/tealeg/xlsx/pull/819/files#diff-736c5868afcf3093a503fb89211daa2dca5ba2b9e2c8fbe5dc105fbbd0e7b963R307
The newly added row index is incorrect, and when reading the data from the row again, it will be empty.


lpxxn:
If saved, the data in the first row will be cleared.

image

tealeg (approved):
+1... good work!

tealeg:
Thank you for your work @lpxxn - much appreciated!

PR #834, opened 2024-12-03
Fix globally available defined names

Sajito:
Defined names can apply globally (i.e. for every sheet in the document). Global defined names have no LocalSheetID attached to them. LocalSheetID=0 means that the defined name only applies to the first sheet. By changing the type to `*int` the information will be preserved correctly.


Sajito:
Still relevant.

PR #838, opened 2025-01-22
Fix hyperlinks in excel

### LINKED ISSUE #836: error when save xlsx file with hyperlink
when i open the new file, office tell me the sheet has invalid xml obj, after i delete the hyperlink cell, it works on.
version github.com/tealeg/xlsx/v3 v3.3.11

Sajito:
The elements inside a worksheet are defined as a sequence, meaning excel expects the elements in a specific order.
Hyperlinks currently are written before sheetData, but must come after sheetData, otherwise it will fail to read them.
Also internal links need to be stored in `location`, which currently get's lost during save.

This fixes #836. Also it adds support for handling internal links in Excel-readable way.

### DIFF
--- a/cell.go
+++ b/cell.go
@@ -376,11 +376,11 @@ func (c *Cell) SetHyperlink(hyperlink string, displayText string, tooltip string
 	h := strings.ToLower(hyperlink)
 	if strings.HasPrefix(h, "http:") || strings.HasPrefix(h, "https://") {
 		c.Hyperlink = Hyperlink{Link: hyperlink}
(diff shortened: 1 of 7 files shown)


cossbow:
I had the same trouble, so I tested the patch myself, and it worked fine except that it didn't automatically turn blue and underline. You need to manually set the style.

tealeg (approved):

Sajito:
@tealeg thanks for merging!

Could you also review and possibly merge #834? It was closed automatically, but is still relevant. I can create a new pr for it, if needed.

PR #839, opened 2025-02-17
Change LocalSheetID to pointer of int

Sajito:
Sorry for creating as new PR, but it's still relevant. I can't reopen the closed pr.

Defined names can apply globally (i.e. for every sheet in the document). Global defined names have no LocalSheetID attached to them. LocalSheetID=0 means that the defined name only applies to the first sheet. By changing the type to `*int` the information will be preserved correctly.


tealeg (approved):
LGTM thanks @Sajito

PR #840, opened 2025-02-17
Handle unsigned ints

adambaratz:
When setting the value of a cell to an unsigned int, it can be handled identically to signed ints.


adambaratz:
Without this change, unsigned ints won't get written correctly with calls to Row.WriteSlice or Row.WriteStruct.

tealeg (approved):
+1 LGTM

PR #846, opened 2025-04-04
Fixes issue when newly added rows are missing from file

### LINKED ISSUE #845: Rows missing from file if sheet name were changed before rows were added
It seems as if #783 only were partially fixed. To fix our current issue, `makeCellKeyRowPrefix` (in row.go) needs to be changed to use `r.Sheet.cellStoreName` instead of `r.Sheet.Name`. I'm guessing that `makeCellKey` (also in row.go), also should be changed to use `cellStoreName`. 

Simple program to reproduce the issue:
```golang
package main

import (
	"fmt"
	"github.com/tealeg/xlsx/v3"
)

func main() {
	wb := xlsx.NewFile()
	sh, _ := wb.AddSheet("My New Sheet")

	setCell := func(row, col int, val string) {
		c, _ := sh.Cell(row, col)
		c.Value = val
	}

	setCell(0, 0, "first row")  // OK
	setCell(1, 0, "second row") // OK
	setCell(2, 0, "third row")  // Missing in file

	// Rename sheet to trigger issue
	sh.Name = "New Sheet Name"

	setCell(3, 0, "fourth row") // Missing in file
	setCell(4, 0, "fifth row")  // Missing in file
	setCell(5, 0, "sixth row")  // Missing in file

	setCell(0, 1, "first row, second column")  // OK since row existed before renaming sheet
	setCell(1, 1, "second row, second column") // OK since row existed before renaming sheet
	setCell(2, 1, "third row, second column")  // Missing in file
	setCell(3, 1, "fourth row, second column") // Missing in file
	setCell(4, 1, "fifth row, second column")  // Missing in file
	setCell(5, 1, "sixth row, second column")  // Missing in file

	err := wb.Save("out.xlsx")
	if err != nil {
		fmt.Printf("Failed to save doc, error: %v", err)
	}
}
```

larsve:
If a sheet were renamed and then new rows were added, the new rows would not be saved in the .xlsx-file.

To fix that, makeCellKeyRowPrefix() and makeCellKey() were updated to use cellStoreName, so they use the same key-prefix as makeRowKey does.

Should fix #845

### DIFF
--- a/row.go
+++ b/row.go
@@ -84,7 +84,7 @@ func (r *Row) PushCell(c *Cell) {
 }
 
 func (r *Row) makeCellKey(colIdx int) string {
-	return fmt.Sprintf("%s:%06d:%06d", r.Sheet.Name, r.num, colIdx)
+	return fmt.Sprintf("%s:%06d:%06d", r.Sheet.cellStoreName, r.num, colIdx)
 }
 
 func (r *Row) key() string {
@@ -92,7 +92,7 @@ func (r *Row) key() string {
 }
 
 func (r *Row) makeCellKeyRowPrefix() string {
-	return fmt.Sprintf("%s:%06d", r.Sheet.Name, r.num)
+	return fmt.Sprintf("%s:%06d", r.Sheet.cellStoreName, r.num)
 }
 
 // GetCell returns the Cell at a given column index, creating it if it doesn't exist.


tealeg (approved):
👍🏻 Thanks very much @larsve

PR #847, opened 2025-04-20
fix: Panic caused by the absence of LastClr

shcabin:
ECMA-376-1:2016   states that lastClr is optional
`<xsd:attribute name="lastClr" type="s:ST_HexColorRGB" use="optional"/>`
and example: 
屏幕截图 2025-04-21 001108

When lastClr does not exist, it will result in empty strings in the theme ("lt1", "dk1" ).

However, I am not sure if returning an empty string is the best option for `themeColor()` in this case.


tealeg:
"Applications shall use the @lastClr attribute to determine the absolute value of the last color used if system colors are not `supported.`" - in any case, if it simply isn't present, there's not much we can actually do about that.

tealeg:
Basically it seems we have two problems here:
-  we're round tripping style information via a mechanism that was designed to only extract a usable colour for on screen presentation, which is actually a fairly unlikely use-case for the library - in general it would be better to preserve the underlying complex colour definition rather than forcing everything to aRGB. 
- we're "resorting" to lastClr because we don't have a definition of the named "system colors". 

Defaulting to the empty string here is likely to have odd results for some people.

theme.go
```diff
@@ -15,7 +15,7 @@ func newTheme(themeXml xlsxTheme) *theme {
 	for _, scheme := range clrSchemes {
 		var rgbColor string
 		if scheme.SysClr != nil {
-			rgbColor = scheme.SysClr.LastClr
+			rgbColor = scheme.SysClr.LastClr // LastClr is optional, so rgbColor may be an empty string
```
tealeg, line 18: Nice comment.  I feel like we really need to make the underlying options for a colour transparent to the public interface of the library.  At the moment all of this code is pointing towards a use case where we need to know a concrete aRGB value for the colour - the only case where you need that is if you're rendering the document somewhere and this seems to be a less likely use case than generating or modifying a document.  In that case we'd be better off just preserving the stylistic information as it is.   Sadly fixing that means creating a v4 and I don't have the time or inclination to tackle it.

tealeg: Given the way we *intend* this code to be used, the right fix for this situation is to define the values for the systemColors and use those values when lastClr is omitted.

theme.go
```diff
@@ -31,7 +31,7 @@ func (t *theme) themeColor(index int64, tint float64) string {
 	baseColor := t.colors[index]
 	if tint == 0 {
 		return "FF" + baseColor
-	} else {
+	} else if baseColor != "" {
```
tealeg, line 34: It's not clear that this won't create unexpected errors.   See my comment above.  The way in v3 would be to define the system colours in a lookup table and insert those values when lastClr is missing.

tealeg (changes requested):
See comments inline.

